PR #47, opened 2026-02-19
Add gatewayAuth field to decouple SecurityPolicy from OIDC client provisioning

tylerpotts:
### Problem                                                                                                                                                                                           
                                                                                                                                                                                                        
 When `auth.enabled: true`, the operator both provisions a Keycloak OIDC client **and** creates an Envoy Gateway SecurityPolicy. These are currently coupled - you get both or neither.                
                                                                                                                                                                                                      
 Apps like Grafana have native OAuth support and need to handle the OIDC flow themselves for user identity, role mapping, and session management. When the SecurityPolicy is active, Envoy intercepts
  the same OAuth callback path that Grafana uses (`/login/generic_oauth`), causing a 401 conflict. Setting `auth.enabled: false` avoids the SecurityPolicy but also skips client provisioning entirely,
  forcing manual Keycloak client setup.

 ### Solution

  Add a `gatewayAuth` boolean field to `AuthConfig` (default: `true`) that controls SecurityPolicy creation independently from client provisioning:

  | `auth.enabled` | `gatewayAuth` | Keycloak client | SecurityPolicy | Use case |
  |---|---|---|---|---|
  | `true` | `true` (default) | Created | Created | App has no native auth (e.g., nginx) |
  | `true` | `false` | Created | **Skipped** | App handles OAuth natively (e.g., Grafana) |
  | `false` | N/A | Skipped | Skipped | No auth needed |

  When `gatewayAuth` is set to `false`:
  - The operator still provisions the Keycloak client and stores credentials in a Secret
  - The application reads the client secret and handles OAuth directly
  - No SecurityPolicy is created, so Envoy passes traffic through without intercepting callbacks

  If an existing NebariApp transitions from `gatewayAuth: true` to `false`, the operator cleans up the previously created SecurityPolicy.

### Changes

  - **`api/v1/nebariapp_types.go`** - Add `GatewayAuth *bool` field to `AuthConfig` with `+kubebuilder:default=true`
  - **`internal/controller/reconcilers/auth/reconciler.go`** - Conditionally skip SecurityPolicy creation when `gatewayAuth: false`, and delete existing SecurityPolicy if transitioning
  - **`config/crd/bases/reconcilers.nebari.dev_nebariapps.yaml`** - Regenerated CRD manifest
  - **`api/v1/zz_generated.deepcopy.go`** - Regenerated deepcopy

### DIFF
--- a/internal/controller/reconcilers/auth/reconciler_test.go
+++ b/internal/controller/reconcilers/auth/reconciler_test.go
@@ -602,7 +602,7 @@ func TestReconcileAuth(t *testing.T) {
 			},
 			existingSecurityPolicy: &egv1alpha1.SecurityPolicy{
 				ObjectMeta: metav1.ObjectMeta{
-					Name:      "test-app-security-policy",
+					Name:      "test-app-security",
 					Namespace: "default",
 				},
 			},


api/v1/nebariapp_types.go
```diff
@@ -164,6 +164,17 @@ type AuthConfig struct {
 	// +optional
 	ProvisionClient *bool `json:"provisionClient,omitempty"`
 
+	// GatewayAuth determines whether the operator should create an Envoy Gateway
+	// SecurityPolicy to enforce authentication at the gateway level.
+	// When true (default), the operator creates a SecurityPolicy that handles
+	// the OIDC flow at the gateway before requests reach the application.
+	// When false, the operator provisions the OIDC client and stores credentials
+	// in a Secret, but does NOT create a SecurityPolicy — the application is
```
dcmcand: This line uses an em dash. Project convention is to avoid em dashes - use a regular dash or rewrite the sentence. This also propagates into the generated CRD description.

Suggestion:
```suggestion
	// in a Secret, but does NOT create a SecurityPolicy - the application is
```

PR #67, opened 2026-03-11
Fix OIDC token exchange TLS failure by using internal token endpoint

tylerpotts:
## Summary

- Fix OAuth flow failure (`CERTIFICATE_VERIFY_FAILED`) when Envoy exchanges the auth code for a token with Keycloak
- Add `GetTokenEndpoint` to the `OIDCProvider` interface so providers can specify an explicit token endpoint
- Keycloak provider returns the internal HTTP token endpoint, generic OIDC provider defers to discovery

## Problem

When a NebariApp has OIDC authentication enabled, the SecurityPolicy's issuer is set to the **internal** Keycloak HTTP URL:
```
http://keycloak-keycloakx-http.keycloak.svc.cluster.local:8080/auth/realms/nebari
```

Envoy fetches the OIDC discovery document from this URL successfully. However, Keycloak has an external frontend URL configured, so the discovery document returns the **external** HTTPS URL for the token endpoint:
```
https://keycloak.nebari.local/auth/realms/nebari/protocol/openid-connect/token
```

When the user completes Keycloak login and gets redirected back with an auth code, Envoy tries to exchange the code for tokens by calling this external HTTPS endpoint. This fails because:
- In local/dev environments: the TLS certificate is self-signed and not trusted by Envoy
- Error: `CERTIFICATE_VERIFY_FAILED` → 401 "OAuth flow failed"

The browser redirect to Keycloak login works fine (that's a browser-side redirect using the external URL), but the server-to-server token exchange fails.

## Solution

Explicitly set the `tokenEndpoint` field on the SecurityPolicy's OIDC provider configuration. This tells Envoy to use the specified URL for token exchange instead of the one discovered from the OIDC configuration document.

For Keycloak, the token endpoint is set to:
```
http://keycloak-keycloakx-http.keycloak.svc.cluster.local:8080/auth/realms/nebari/protocol/openid-connect/token
```

This keeps the server-to-server token exchange inside the cluster over HTTP, avoiding TLS verification entirely.

### Why this also benefits cloud deployments

Even when TLS certificates are valid (e.g., Let's Encrypt on cloud), using the internal endpoint is better because:
- Token exchange stays within the cluster (lower latency)
- No hairpin routing through the external load balancer
- No dependency on external DNS resolution for server-to-server calls

### Changes

| File | Change |
|------|--------|
| `providers/provider.go` | Add `GetTokenEndpoint` to `OIDCProvider` interface |
| `providers/keycloak.go` | Implement `GetTokenEndpoint` returning internal HTTP URL |
| `providers/generic_oidc.go` | Implement `GetTokenEndpoint` returning empty (use discovery) |
| `auth/reconciler.go` | Set `TokenEndpoint` on SecurityPolicy when provider returns one |
| `auth/reconciler_test.go` | Add test cases for both explicit and discovery token endpoints |

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./internal/controller/reconcilers/auth/...` passes (all existing + 2 new tests)
- [ ] Deploy operator to local kind cluster and verify OIDC login flow completes successfully
- [ ] Verify SecurityPolicy shows `tokenEndpoint` set to internal URL
- [ ] Verify generic-oidc provider does not set tokenEndpoint (falls back to discovery)

### DIFF
--- a/internal/controller/reconcilers/auth/providers/generic_oidc.go
+++ b/internal/controller/reconcilers/auth/providers/generic_oidc.go
@@ -37,6 +37,12 @@ func (p *GenericOIDCProvider) GetIssuerURL(ctx context.Context, nebariApp *appsv
 	return nebariApp.Spec.Auth.IssuerURL, nil
 }
 
+// GetTokenEndpoint returns empty string for generic OIDC providers.
+// The token endpoint will be discovered from the OIDC provider's well-known configuration.
+func (p *GenericOIDCProvider) GetTokenEndpoint(ctx context.Context, nebariApp *appsv1.NebariApp) string {
+	return ""
+}
+
 // GetExternalIssuerURL returns the same URL as GetIssuerURL for generic OIDC.
 // Generic OIDC issuer URLs are already externally routable by definition.
 func (p *GenericOIDCProvider) GetExternalIssuerURL(ctx context.Context, nebariApp *appsv1.NebariApp) (string, error) {
--- a/internal/controller/reconcilers/auth/providers/generic_oidc.go
+++ b/internal/controller/reconcilers/auth/providers/generic_oidc.go
@@ -37,10 +37,11 @@ func (p *GenericOIDCProvider) GetIssuerURL(ctx context.Context, nebariApp *appsv
 	return nebariApp.Spec.Auth.IssuerURL, nil
 }
 
-// GetTokenEndpoint returns empty string for generic OIDC providers.
-// The token endpoint will be discovered from the OIDC provider's well-known configuration.
-func (p *GenericOIDCProvider) GetTokenEndpoint(ctx context.Context, nebariApp *appsv1.NebariApp) string {
-	return ""
+// GetEndpointOverrides returns empty overrides for generic OIDC providers.
+// Generic OIDC issuers are externally routable by definition; Envoy can reach
+// the discovered endpoints directly, so no overrides are needed.
+func (p *GenericOIDCProvider) GetEndpointOverrides(_ context.Context, _ *appsv1.NebariApp) (OIDCEndpointOverrides, error) {
+	return OIDCEndpointOverrides{}, nil
 }
 
 // GetExternalIssuerURL returns the same URL as GetIssuerURL for generic OIDC.
--- a/internal/controller/reconcilers/auth/providers/keycloak.go
+++ b/internal/controller/reconcilers/auth/providers/keycloak.go
@@ -48,7 +48,7 @@ type KeycloakProvider struct {
 }
 
 // internalRealmURL returns the base internal cluster URL for the Keycloak realm.
-// This is used by both GetIssuerURL and GetTokenEndpoint to avoid duplication.
+// This is used by both GetIssuerURL and GetEndpointOverrides to avoid duplication.
 func (p *KeycloakProvider) internalRealmURL() string {
 	return fmt.Sprintf("http://%s.%s.svc.cluster.local:%d%s/realms/%s",
 		p.Config.IssuerServiceName,
@@ -64,12 +64,19 @@ func (p *KeycloakProvider) GetIssuerURL(ctx context.Context, nebariApp *appsv1.N
 	return p.internalRealmURL(), nil
 }
 
-// GetTokenEndpoint returns the internal cluster token endpoint URL for Keycloak.
-// This avoids Envoy using the external HTTPS token endpoint from the OIDC discovery
+// GetEndpointOverrides returns internal cluster URLs for Keycloak's OIDC endpoints.
+// This avoids Envoy using the external HTTPS endpoints from the OIDC discovery
 // document, which may use a certificate not trusted by Envoy (e.g., self-signed).
-func (p *KeycloakProvider) GetTokenEndpoint(ctx context.Context, nebariApp *appsv1.NebariApp) string {
-	return p.internalRealmURL() + "/protocol/openid-connect/token"
+func (p *KeycloakProvider) GetEndpointOverrides(_ context.Context, _ *appsv1.NebariApp) (OIDCEndpointOverrides, error) {
+	base := p.internalRealmURL() + "/protocol/openid-connect"
+	return OIDCEndpointOverrides{
+		Token:         ptrTo(base + "/token"),
+		Authorization: ptrTo(base + "/auth"),
+		EndSession:    ptrTo(base + "/logout"),
+	}, nil
 }
+
+func ptrTo(s string) *string { return &s }
 
 // GetExternalIssuerURL returns the publicly routable Keycloak issuer URL.
 func (p *KeycloakProvider) GetExternalIssuerURL(ctx context.Context, nebariApp *appsv1.NebariApp) (string, error) {
--- a/internal/controller/reconcilers/auth/providers/keycloak.go
+++ b/internal/controller/reconcilers/auth/providers/keycloak.go
@@ -47,18 +47,28 @@ type KeycloakProvider struct {
 	Config config.KeycloakConfig
 }
 
-// GetIssuerURL returns the internal cluster URL for the Keycloak realm.
-// Envoy uses this to fetch OIDC configuration from within the cluster.
-func (p *KeycloakProvider) GetIssuerURL(ctx context.Context, nebariApp *appsv1.NebariApp) (string, error) {
-	realm := p.Config.Realm
-	// Use internal cluster DNS for Envoy to fetch OIDC config
-	// All components are now configurable via environment variables
+// internalRealmURL returns the base internal cluster URL for the Keycloak realm.
+// This is used by both GetIssuerURL and GetTokenEndpoint to avoid duplication.
+func (p *KeycloakProvider) internalRealmURL() string {
 	return fmt.Sprintf("http://%s.%s.svc.cluster.local:%d%s/realms/%s",
 		p.Config.IssuerServiceName,
 		p.Config.IssuerServiceNamespace,
 		p.Config.IssuerServicePort,
 		p.Config.IssuerContextPath,
-		realm), nil
+		p.Config.Realm)
+}
+
+// GetIssuerURL returns the internal cluster URL for the Keycloak realm.
+// Envoy uses this to fetch OIDC configuration from within the cluster.
+func (p *KeycloakProvider) GetIssuerURL(ctx context.Context, nebariApp *appsv1.NebariApp) (string, error) {
+	return p.internalRealmURL(), nil
+}
+
+// GetTokenEndpoint returns the internal cluster token endpoint URL for Keycloak.
+// This avoids Envoy using the external HTTPS token endpoint from the OIDC discovery
+// document, which may use a certificate not trusted by Envoy (e.g., self-signed).
+func (p *KeycloakProvider) GetTokenEndpoint(ctx context.Context, nebariApp *appsv1.NebariApp) string {
+	return p.internalRealmURL() + "/protocol/openid-connect/token"
 }
 
 // GetExternalIssuerURL returns the publicly routable Keycloak issuer URL.
--- a/internal/controller/reconcilers/auth/providers/keycloak.go
+++ b/internal/controller/reconcilers/auth/providers/keycloak.go
@@ -33,6 +33,7 @@ import (
 	"github.com/nebari-dev/nebari-operator/internal/config"
 	"github.com/nebari-dev/nebari-operator/internal/controller/utils/constants"
 	"github.com/nebari-dev/nebari-operator/internal/controller/utils/naming"
+	"github.com/nebari-dev/nebari-operator/internal/controller/utils/ptr"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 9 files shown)


dcmcand (changes requested):
This PR has a lot of good work in it - the token endpoint fix is well-motivated, the group sync and client-level protocol mapper changes are solid features. But there are a couple things to address before merge.

**Branch divergence**: GitHub shows this PR as unmergeable (`dirty`). The branch diverged from main before several significant features landed (SPA client, landing page discovery, cross-namespace services, public routes, gateway listener conflict detection, e2e parallelization, default context path change, health check port). A rebase onto current main is needed, and care should be taken during the rebase since `keycloak.go`, `reconciler.go`, and the CRD types all have conflicts.

The code itself is well-structured. Inline comments below cover the specific items worth addressing.

viniciusdc (commented):
## Tested on live cluster ✓

Deployed the `fix-oidc-token-endpoint-internal` image to our Hetzner cluster (k3s + Envoy Gateway) and validated the full OIDC auth flow end-to-end.

### What was checked

**SecurityPolicies** — all NebariApps reconciled cleanly with the internal token endpoint set:
```
http://keycloak-keycloakx-http.keycloak.svc.cluster.local:8080/realms/nebari/protocol/openid-connect/token
```

**Envoy access logs** — a real token exchange captured at `13:09:08`:
```json
{
  "method": "POST",
  "x-envoy-origin-path": "/realms/nebari/protocol/openid-connect/token",
  "user-agent": "Go-http-client/1.1",
  "response_code": 200,
  "bytes_sent": 3644,
  "upstream_host": "10.244.0.18:8080",
  "upstream_transport_failure_reason": null
}
```

Envoy's HTTP client (`Go-http-client/1.1`) POSTed to the token endpoint, got a **200** with a full token payload (3644 bytes), routed directly to the Keycloak pod IP over HTTP — no TLS involved. Zero `CERTIFICATE_VERIFY_FAILED` errors in the logs.

**In-cluster reachability** — confirmed the internal token endpoint responds correctly from within the cluster (HTTP 401 with no credentials, as expected).

### Before vs after

| | Before | After |
|---|---|---|
| Token endpoint used | External HTTPS (from OIDC discovery doc) | Internal HTTP (explicitly set on SecurityPolicy) |
| TLS verification | Failed — self-signed cert not trusted by Envoy | Not needed — stays inside cluster |
| Token exchange result | `CERTIFICATE_VERIFY_FAILED` → 401 | 200 with full token response |

dcmcand (commented):
Thanks for the thorough writeup - the internal-fetch / external-discovered-endpoint asymmetry explanation in the PR description is excellent, and the `internalRealmURL()` refactor is exactly the right shape. The test axes for `GetTokenEndpoint` mirror the existing `GetIssuerURL` cases, which is good coverage discipline.

No design-gating blockers. A few questions and one cosmetic blocker around a misplaced godoc comment. Architectural question worth discussing: token is one of three overridable OIDC endpoints in Envoy Gateway (authorization and end-session can hit the same external-HTTPS-from-discovery problem), so it may be worth generalizing the interface now rather than growing it three times. Inline comments below.

internal/controller/reconcilers/auth/providers/provider.go
```diff
@@ -31,6 +31,12 @@ type OIDCProvider interface {
 	// The URL should be accessible from within the cluster (internal DNS).
 	GetIssuerURL(ctx context.Context, nebariApp *appsv1.NebariApp) (string, error)
 
+	// GetTokenEndpoint returns an explicit token endpoint URL if needed.
+	// This is used when the OIDC discovery document returns an external URL
+	// that is not reachable from within the cluster (e.g., uses HTTPS with
+	// a certificate not trusted by Envoy). Returns empty string to use discovery.
+	GetTokenEndpoint(ctx context.Context, nebariApp *appsv1.NebariApp) string
```
dcmcand: **Question (architecture):** Envoy Gateway's `OIDCProvider` type exposes three overridable endpoints - `AuthorizationEndpoint`, `TokenEndpoint`, and `EndSessionEndpoint`. All three can suffer the same external-HTTPS-from-discovery issue when Keycloak has a frontend URL configured (browser-facing authorization endpoint in particular). If we later need to override the other two, this pattern will grow the interface N times for the same underlying concept.

Worth considering a single method returning a struct of overrides now:

```go
type OIDCEndpointOverrides struct {
    Token, Authorization, EndSession *string
}
GetEndpointOverrides(ctx, nebariApp) OIDCEndpointOverrides
```

Reconciler then copies any non-nil fields onto `oidcProvider`. Future overrides added without interface churn and without touching every provider.

dcmcand: **Question (consistency):** Every other URL-producing method on this interface returns `(string, error)` - `GetIssuerURL`, `GetExternalIssuerURL`. Neither of the current `GetTokenEndpoint` implementations can fail, but `GetExternalIssuerURL` has the same property in generic OIDC and still returns `error` for consistency.

Divergent signatures nudge future maintainers either into breaking the signature later (e.g., if a provider discovers the token endpoint lazily) or into skipping error reporting. Suggest `GetTokenEndpoint(ctx, nebariApp) (string, error)`; current impls return `"", nil` / `url, nil`.

dcmcand: **Question:** Neither current implementation uses `ctx` or `nebariApp`. Intentional forward-looking signature (future provider might need per-request context or per-app data), or could this be `GetTokenEndpoint() string`? Worth an explicit decision documented in the interface comment.

internal/controller/reconcilers/auth/providers/provider.go
```diff
@@ -31,6 +31,12 @@ type OIDCProvider interface {
 	// The URL should be accessible from within the cluster (internal DNS).
 	GetIssuerURL(ctx context.Context, nebariApp *appsv1.NebariApp) (string, error)
 
+	// GetTokenEndpoint returns an explicit token endpoint URL if needed.
```
dcmcand: **Stylistic:** The contract is underspecified - a future provider implementer reading only the interface wouldn't know what a valid non-empty return must look like. Consider tightening to something like: "If non-empty, the returned value must be a complete absolute URL reachable from within the cluster."

internal/controller/reconcilers/auth/providers/keycloak.go
```diff
@@ -47,18 +47,28 @@ type KeycloakProvider struct {
...
 
-// GetIssuerURL returns the internal cluster URL for the Keycloak realm.
-// Envoy uses this to fetch OIDC configuration from within the cluster.
-func (p *KeycloakProvider) GetIssuerURL(ctx context.Context, nebariApp *appsv1.NebariApp) (string, error) {
-	realm := p.Config.Realm
-	// Use internal cluster DNS for Envoy to fetch OIDC config
-	// All components are now configurable via environment variables
+// internalRealmURL returns the base internal cluster URL for the Keycloak realm.
+// This is used by both GetIssuerURL and GetTokenEndpoint to avoid duplication.
+func (p *KeycloakProvider) internalRealmURL() string {
```
dcmcand, line 53: Nice refactor - having `GetIssuerURL` and `GetTokenEndpoint` share one URL builder keeps them structurally consistent and avoids the duplicated `fmt.Sprintf` that would otherwise show up.

internal/controller/reconcilers/auth/providers/keycloak.go
```diff
@@ -47,18 +47,28 @@ type KeycloakProvider struct {
...
+// GetIssuerURL returns the internal cluster URL for the Keycloak realm.
+// Envoy uses this to fetch OIDC configuration from within the cluster.
+func (p *KeycloakProvider) GetIssuerURL(ctx context.Context, nebariApp *appsv1.NebariApp) (string, error) {
+	return p.internalRealmURL(), nil
+}
+
+// GetTokenEndpoint returns the internal cluster token endpoint URL for Keycloak.
+// This avoids Envoy using the external HTTPS token endpoint from the OIDC discovery
+// document, which may use a certificate not trusted by Envoy (e.g., self-signed).
+func (p *KeycloakProvider) GetTokenEndpoint(ctx context.Context, nebariApp *appsv1.NebariApp) string {
```
dcmcand: **Stylistic:** Consider naming unused params `_`:

```go
func (p *KeycloakProvider) GetTokenEndpoint(_ context.Context, _ *appsv1.NebariApp) string
```

Signals intent and quiets unused-param linters.

internal/controller/reconcilers/auth/providers/generic_oidc.go
```diff
@@ -37,6 +37,12 @@ func (p *GenericOIDCProvider) GetIssuerURL(ctx context.Context, nebariApp *appsv
 	return nebariApp.Spec.Auth.IssuerURL, nil
 }
 
+// GetTokenEndpoint returns empty string for generic OIDC providers.
+// The token endpoint will be discovered from the OIDC provider's well-known configuration.
+func (p *GenericOIDCProvider) GetTokenEndpoint(ctx context.Context, nebariApp *appsv1.NebariApp) string {
```
dcmcand: **Minor (doc coherence):** The existing comment on `GetExternalIssuerURL` establishes the invariant "Generic OIDC issuer URLs are already externally routable by definition." The new `GetTokenEndpoint` comment only says "will be discovered" without restating why the override is unnecessary. A future contributor could add a `discoverTokenEndpoint` helper here to "be helpful" and silently violate the invariant.

Suggest: "Generic OIDC issuers are externally routable by definition; Envoy can reach the discovered token endpoint directly, so no override is needed."

Same `_` naming suggestion as the keycloak impl.

internal/controller/reconcilers/auth/reconciler.go
```diff
@@ -443,10 +443,20 @@ func (r *AuthReconciler) buildSecurityPolicySpec(ctx context.Context, nebariApp
...
+	oidcProvider := egv1alpha1.OIDCProvider{
+		Issuer: issuerURL,
+	}
+
+	// Set explicit token endpoint if the provider returns one.
+	// This ensures Envoy uses the internal cluster URL for the token exchange
+	// instead of the external URL from the OIDC discovery document, which may
+	// use a TLS certificate not trusted by Envoy.
+	if tokenEndpoint := provider.GetTokenEndpoint(ctx, nebariApp); tokenEndpoint != "" {
+		oidcProvider.TokenEndpoint = ptrTo(tokenEndpoint)
```
dcmcand: **Question (observability):** Other auth outcomes (Provisioned, Configured, ValidationFailed) emit events, but this silently overrides discovery. In cloud deployments where discovery would also have worked, operators debugging the flow would have no signal that an explicit endpoint was set.

Consider a debug-level log or a field on the existing `Configured` event noting the token endpoint override was applied.

internal/controller/reconcilers/auth/reconciler_test.go
```diff
@@ -36,13 +36,27 @@ import (
 )
 
 // boolPtr returns a pointer to a bool value
```
dcmcand: **Blocker (cosmetic):** `verifyTokenEndpoint` was inserted between the `// boolPtr returns a pointer to a bool value` comment and `func boolPtr`, so that comment now documents `verifyTokenEndpoint` instead. `go doc` will show the wrong text, and doc-comment linters (revive, godot) will flag it.

Fix: move `verifyTokenEndpoint` below `boolPtr`, or move the `boolPtr` comment down to sit directly above `func boolPtr`.

internal/controller/reconcilers/auth/reconciler_test.go
```diff
@@ -36,13 +36,27 @@ import (
...
+// verifyTokenEndpoint checks that the SecurityPolicy's token endpoint matches expectations.
+func verifyTokenEndpoint(t *testing.T, sp *egv1alpha1.SecurityPolicy, expected string) {
```
dcmcand: Good use of `t.Helper()` in this helper - failures will point to the call site rather than inside the helper. Also nice that it handles both the nil and non-nil cases.

internal/controller/reconcilers/auth/reconciler_test.go
```diff
@@ -972,6 +1056,8 @@ func TestReconcileAuth(t *testing.T) {
 				if shouldEnforceAtGateway(tt.nebariApp.Spec.Auth) {
 					if err != nil {
 						t.Errorf("expected SecurityPolicy to be created, got error: %v", err)
+					} else if securityPolicy.Spec.OIDC != nil {
+						verifyTokenEndpoint(t, securityPolicy, tt.provider.GetTokenEndpoint(context.Background(), tt.nebariApp))
```
dcmcand: **Question:** This calls `tt.provider.GetTokenEndpoint(...)` a second time to compute the expected value. That means a bug where the reconciler sets `TokenEndpoint` to something different from what the provider returned would not be caught - the test would just compare the provider's return value to itself (via the reconciler). Literal expected URLs in the test table (the way `expectedURL` is done in `TestKeycloakProvider_GetTokenEndpoint`) would make the assertion independent of the mock.

viniciusdc (commented):
Thanks for the thorough review, @dcmcand! Really appreciate the architectural thinking here. Addressed everything in 4bfcd73:

### Blocker
- **Misplaced godoc comment** — Fixed. Moved `boolPtr` above `verifyTokenEndpoint` so the comment sits on the right function.

### Suggestions adopted
- **Struct-based overrides (Q1)** — Agreed this was the right call. Replaced `GetTokenEndpoint` with `GetEndpointOverrides() (OIDCEndpointOverrides, error)` returning a struct with `Token`, `Authorization`, and `EndSession` pointer fields. Keycloak now overrides all three with internal cluster URLs — which actually makes deployments more robust since the authorization and end-session endpoints had the same external-HTTPS problem.
- **Signature consistency (Q2)** — The new method returns `(OIDCEndpointOverrides, error)`, matching the `(string, error)` convention of `GetIssuerURL` and `GetExternalIssuerURL`.
- **Unused params (Q3 + stylistic)** — Kept `ctx` and `nebariApp` in the interface signature for forward compatibility (a future provider might need them), but used `_` in current implementations to signal intent and quiet linters.
- **Interface doc (stylistic)** — Added: "Any non-nil field must be a complete absolute URL reachable from within the cluster."
- **Generic OIDC doc (minor)** — Updated to: "Generic OIDC issuers are externally routable by definition; Envoy can reach the discovered endpoints directly, so no overrides are needed."
- **Observability (Q4)** — Added per-endpoint `Info`-level log when each override is applied, e.g. `Overriding OIDC endpoint from discovery endpoint=token url=http://...`.
- **Test independence (Q5)** — Added `expectedOverrides` field with literal URLs to the reconciler test table, so assertions compare against hardcoded values instead of re-calling the mock.

viniciusdc (commented):
A few more follow-up commits on top of the previous round:

- Extracted the duplicated `ptrTo` helpers (one in `keycloak.go`, one in `reconciler.go`) into a shared `internal/controller/utils/ptr` package with a generic `ptr.To[T]`. Also added a test case for `GetEndpointOverrides` returning an error — the error-wrapping path in `buildSecurityPolicySpec` wasn't covered before.

- Replaced the local `boolPtr`/`strPtr` helpers in `reconciler_test.go` with `ptr.To` since we now have the shared utility. Keeps things consistent across production and test code.

All tests passing, no API/CRD changes — purely internal cleanup.

viniciusdc (approved):

PR #99, opened 2026-04-08
fix: repair goreleaser hooks and clean up release docs

viniciusdc:
## Summary

Fixes two bugs found during the `v0.1.0-alpha.16` release postmortem and cleans up inconsistencies in the release documentation.

### Bugs fixed

**1. GoReleaser `before.hooks` failing on every release** (root cause of missing Go binaries in the release)

GoReleaser v2 does not run hook strings through a shell — it parses them as `argv`. This means `|| true` was passed as a literal argument to `git` rather than evaluated as a shell operator, causing the hook to abort immediately. Wrapped with `sh -c`:

```yaml
# before (broken)
- git checkout -- dist/ || true

# after (fixed)
- sh -c "git checkout -- dist/ 2>/dev/null || true"
```

**2. Helm OCI path collides with Docker image tag**

The `helm-repository` release workflow pushes charts to `oci://quay.io/nebari`, which resolves to `quay.io/nebari/nebari-operator:0.1.0-alpha.16` — **the same tag as the Docker container image**. This silently overwrites the container image with a Helm OCI artifact (~12 KB), causing the `stat /manager: no such file or directory` crash on the cluster. A separate issue has been filed in nebari-dev/helm-repository to fix the push path. This PR updates all install examples to reference the correct future path (`oci://quay.io/nebari/charts/nebari-operator`).

### Doc fixes

- Remove duplicate "Step 4" section in `release-process.md`
- Fix step ordering in `release-checklist.md`: the release workflow triggers on `release: published`, not on tag push — so the GitHub Release must be created **before** monitoring the workflow, not after
- Update Helm install examples in both doc files to use the OCI path

### Misc

- Remove unreachable `sed`/`rm` lines after `exit 1` in the `prepare-release` Makefile target

## Test plan

- [ ] Trigger a new release and confirm GoReleaser completes without the `hook failed` error
- [ ] Confirm Go binaries are attached to the release
- [ ] Confirm Docker image at `quay.io/nebari/nebari-operator:<tag>` remains a container image after the helm-repository sync runs

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -5,9 +5,11 @@ version: 2
 
 before:
   hooks:
-    # Clean up ignored files that may be in the git tree from release commits
-    - git checkout -- dist/ || true
-    - git clean -fd dist/ || true
+    # Clean up ignored files that may be in the git tree from release commits.
+    # Note: hooks are NOT run through a shell by default in GoReleaser v2, so
+    # shell operators like || must be wrapped explicitly with sh -c.
+    - sh -c "git checkout -- dist/ 2>/dev/null || true"
+    - sh -c "git clean -fd dist/ 2>/dev/null || true"
     # Ensure dependencies are up to date
     - go mod tidy
     # Generate code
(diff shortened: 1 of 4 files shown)


viniciusdc:
Closing this in favor of #101, which takes a more comprehensive approach by consolidating Docker builds into GoReleaser `dockers_v2` and removing the broken hooks entirely — a cleaner fix than the `sh -c` wrapper here.

The non-overlapping changes from this PR (release docs fixes, Makefile cleanup, and ginkgo-to-go-test migration) will be moved to a smaller follow-up PR.

PR #106, opened 2026-04-14
fix: propagate SetControllerReference error in routing reconciler

### LINKED ISSUE #31: bug: Ignored SetControllerReference error in routing reconciler
## Summary

In the routing reconciler, the error from `controllerutil.SetControllerReference()` is silently discarded when building the HTTPRoute. This means if the owner reference fails to be set, the HTTPRoute will be created without an owner reference, preventing garbage collection when the NebariApp is deleted.

## Problem

In `internal/controller/reconcilers/routing/httproute.go`:
```go
// httproute.go:195 - error silently discarded
_ = controllerutil.SetControllerReference(nebariApp, route, r.Scheme)
```

Compare with the correct handling in the auth reconciler (`reconciler.go:216-217`):
```go
if err := controllerutil.SetControllerReference(nebariApp, securityPolicy, r.Scheme); err != nil {
    return fmt.Errorf("failed to set controller reference: %w", err)
}
```

### When Can This Fail?
- The NebariApp scheme is not registered (unlikely but possible in tests or misconfiguration)
- The HTTPRoute already has a different controller owner (cross-namespace ownership edge case)
- The NebariApp UID is empty (in-memory object not yet persisted)

### Impact
- HTTPRoute created without owner reference
- When NebariApp is deleted, the HTTPRoute becomes an orphan and is never cleaned up
- `CleanupHTTPRoute` does explicitly delete it, but if cleanup fails (e.g., controller crashes mid-cleanup), the orphan persists

## Expected Behavior

The error should be propagated to prevent creating an HTTPRoute without proper ownership:
```go
if err := controllerutil.SetControllerReference(nebariApp, route, r.Scheme); err != nil {
    return nil, fmt.Errorf("failed to set controller reference on HTTPRoute: %w", err)
}
```

Note: `buildHTTPRoute` currently returns `*gatewayv1.HTTPRoute`. It would need to return `(*gatewayv1.HTTPRoute, error)` to propagate the error, and `ReconcileRouting` would need to handle it.

## Files

- `internal/controller/reconcilers/routing/httproute.go` - `buildHTTPRoute()` method (line 195)
- `internal/controller/reconcilers/routing/httproute.go` - `ReconcileRouting()` method (line 47) - caller

## Definition of Done

- [ ] `buildHTTPRoute` returns an error when `SetControllerReference` fails
- [ ] `ReconcileRouting` handles the error from `buildHTTPRoute`
- [ ] Unit test confirms error propagation when `SetControllerReference` fails

viniciusdc:
## Summary

- Change `buildHTTPRoute` and `buildPublicHTTPRoute` to return `(*gatewayv1.HTTPRoute, error)` instead of silently discarding the `SetControllerReference` error
- Propagate the error through `ReconcileRouting` and `ReconcilePublicRoute`
- Set `RoutingReady=False` condition on both error paths so the failure is observable on the resource
- Add `TestBuildHTTPRoute_SetControllerReferenceError`, `TestBuildPublicHTTPRoute_SetControllerReferenceError`, and `TestReconcileRouting_BuildError` to confirm error propagation and condition wiring

## Problem

Both builder functions were using a blank assignment to discard the error:

```go
_ = controllerutil.SetControllerReference(nebariApp, route, r.Scheme)
```

If this call fails, the HTTPRoute is created without an owner reference. When the NebariApp is later deleted, the orphaned HTTPRoute is never garbage collected.

## Changes

| File | Change |
|---|---|
| `routing/httproute.go` | `buildHTTPRoute` and `buildPublicHTTPRoute` now return `error`; callers handle it and set condition |
| `routing/httproute_test.go` | All call sites updated; three new error-path tests added |

### How to test locally

1. Run the new tests directly:
   ```bash
   go test ./internal/controller/reconcilers/routing/... -v -run "SetControllerReference|BuildError"
   ```

2. Run the full routing suite to check for regressions:
   ```bash
   go test ./internal/controller/reconcilers/routing/...
   ```

3. To verify the orphan-prevention behaviour on a live cluster, delete a `NebariApp` and confirm its `HTTPRoute` is also removed:
   ```bash
   kubectl delete nebariapp <name> -n <namespace>
   kubectl get httproute -n <namespace>  # should be gone
   ```
   Before this fix, a failure in `SetControllerReference` would leave the `HTTPRoute` behind even after the `NebariApp` was deleted.

### What a regression looks like

If the error handling is removed or swallowed, `TestReconcileRouting_BuildError` will fail — it explicitly asserts that `ReconcileRouting` returns a non-nil error **and** sets `RoutingReady=False` when the build step fails.

Closes #31

### DIFF
--- a/internal/controller/reconcilers/routing/httproute.go
+++ b/internal/controller/reconcilers/routing/httproute.go
@@ -64,7 +64,13 @@ func (r *RoutingReconciler) ReconcileRouting(ctx context.Context, nebariApp *app
 	}
 
 	// Generate desired HTTPRoute
-	desiredRoute := r.buildHTTPRoute(nebariApp, gatewayName, tlsListenerName)
+	desiredRoute, err := r.buildHTTPRoute(nebariApp, gatewayName, tlsListenerName)
+	if err != nil {
+		logger.Error(err, "Failed to build HTTPRoute")
+		conditions.SetCondition(nebariApp, appsv1.ConditionTypeRoutingReady, metav1.ConditionFalse,
+			"BuildFailed", fmt.Sprintf("Failed to build HTTPRoute: %v", err))
+		return err
+	}
 
 	// Check if HTTPRoute already exists
 	existingRoute := &gatewayv1.HTTPRoute{}
@@ -73,7 +79,7 @@ func (r *RoutingReconciler) ReconcileRouting(ctx context.Context, nebariApp *app
 		Namespace: desiredRoute.Namespace,
 	}
 
-	err := r.Client.Get(ctx, routeKey, existingRoute)
+	err = r.Client.Get(ctx, routeKey, existingRoute)
 	if err != nil {
 		if errors.IsNotFound(err) {
 			// Create new HTTPRoute
@@ -153,7 +159,7 @@ func (r *RoutingReconciler) CleanupHTTPRoute(ctx context.Context, nebariApp *app
 // buildHTTPRoute generates an HTTPRoute resource from NebariApp spec.
 // tlsListenerName overrides the default "https" section name when TLS is enabled
 // and a per-app TLS listener has been created by the TLS reconciler.
-func (r *RoutingReconciler) buildHTTPRoute(nebariApp *appsv1.NebariApp, gatewayName string, tlsListenerName string) *gatewayv1.HTTPRoute {
+func (r *RoutingReconciler) buildHTTPRoute(nebariApp *appsv1.NebariApp, gatewayName string, tlsListenerName string) (*gatewayv1.HTTPRoute, error) {
 	routeName := naming.HTTPRouteName(nebariApp)
 	namespace := gatewayv1.Namespace(constants.GatewayNamespace)
 
@@ -208,9 +214,11 @@ func (r *RoutingReconciler) buildHTTPRoute(nebariApp *appsv1.NebariApp, gatewayN
 	}
 
 	// Set owner reference for garbage collection
-	_ = controllerutil.SetControllerReference(nebariApp, route, r.Scheme)
+	if err := controllerutil.SetControllerReference(nebariApp, route, r.Scheme); err != nil {
+		return nil, fmt.Errorf("failed to set controller reference on HTTPRoute: %w", err)
+	}
 
-	return route
+	return route, nil
 }
 
 // buildHTTPRouteRules generates HTTPRoute rules based on NebariApp routes
@@ -302,7 +310,13 @@ func (r *RoutingReconciler) ReconcilePublicRoute(ctx context.Context, nebariApp 
 	logger.Info("Reconciling public route", "gateway", gatewayName, "hostname", nebariApp.Spec.Hostname,
 		"publicRoutes", nebariApp.Spec.Routing.PublicRoutes)
 
-	desiredRoute := r.buildPublicHTTPRoute(nebariApp, gatewayName, tlsListenerName)
+	desiredRoute, err := r.buildPublicHTTPRoute(nebariApp, gatewayName, tlsListenerName)
+	if err != nil {
+		logger.Error(err, "Failed to build public HTTPRoute")
+		conditions.SetCondition(nebariApp, appsv1.ConditionTypeRoutingReady, metav1.ConditionFalse,
+			"BuildFailed", fmt.Sprintf("Failed to build public HTTPRoute: %v", err))
+		return err
+	}
 
 	existingRoute := &gatewayv1.HTTPRoute{}
 	routeKey := client.ObjectKey{
@@ -310,7 +324,7 @@ func (r *RoutingReconciler) ReconcilePublicRoute(ctx context.Context, nebariApp 
 		Namespace: desiredRoute.Namespace,
 	}
 
-	err := r.Client.Get(ctx, routeKey, existingRoute)
+	err = r.Client.Get(ctx, routeKey, existingRoute)
 	if err != nil {
 		if errors.IsNotFound(err) {
 			if err := r.Client.Create(ctx, desiredRoute); err != nil {
@@ -375,7 +389,7 @@ func (r *RoutingReconciler) CleanupPublicHTTPRoute(ctx context.Context, nebariAp
 
 // buildPublicHTTPRoute generates an HTTPRoute for public routes that bypass OIDC authentication.
 // This route is separate from the main route so the SecurityPolicy only targets the main route.
-func (r *RoutingReconciler) buildPublicHTTPRoute(nebariApp *appsv1.NebariApp, gatewayName string, tlsListenerName string) *gatewayv1.HTTPRoute {
+func (r *RoutingReconciler) buildPublicHTTPRoute(nebariApp *appsv1.NebariApp, gatewayName string, tlsListenerName string) (*gatewayv1.HTTPRoute, error) {
 	routeName := naming.PublicHTTPRouteName(nebariApp)
 	namespace := gatewayv1.Namespace(constants.GatewayNamespace)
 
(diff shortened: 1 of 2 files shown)


internal/controller/reconcilers/routing/httproute.go
```diff
@@ -208,9 +214,11 @@ func (r *RoutingReconciler) buildHTTPRoute(nebariApp *appsv1.NebariApp, gatewayN
 	}
 
 	// Set owner reference for garbage collection
-	_ = controllerutil.SetControllerReference(nebariApp, route, r.Scheme)
+	if err := controllerutil.SetControllerReference(nebariApp, route, r.Scheme); err != nil {
+		return nil, fmt.Errorf("failed to set controller reference on HTTPRoute: %w", err)
```
dcmcand, line 218: Nice - wrapping with `%w` here preserves the error chain so callers up the stack can `errors.Is` / `errors.As` if they ever need to. Same for line 459 in `buildPublicHTTPRoute`.

internal/controller/reconcilers/routing/httproute.go
```diff
@@ -64,7 +64,13 @@ func (r *RoutingReconciler) ReconcileRouting(ctx context.Context, nebariApp *app
 	}
 
 	// Generate desired HTTPRoute
-	desiredRoute := r.buildHTTPRoute(nebariApp, gatewayName, tlsListenerName)
+	desiredRoute, err := r.buildHTTPRoute(nebariApp, gatewayName, tlsListenerName)
+	if err != nil {
+		logger.Error(err, "Failed to build HTTPRoute")
+		conditions.SetCondition(nebariApp, appsv1.ConditionTypeRoutingReady, metav1.ConditionFalse,
+			"BuildFailed", fmt.Sprintf("Failed to build HTTPRoute: %v", err))
```
dcmcand, line 71: **Stylistic / question:** the new reason `"BuildFailed"` is an inline literal, while related reasons like `EventReasonGatewayNotFound` are defined as constants in `api/v1/nebariapp_types.go`. The file already mixes both styles (`"CreationFailed"`, `"UpdateFailed"` are also literals), so this is locally consistent, but the API surface won't expose it. Worth adding `EventReasonHTTPRouteBuildFailed` (and using it at line 317 too) so the reason is discoverable from the API package. Non-blocking.

internal/controller/reconcilers/routing/httproute.go
```diff
@@ -64,7 +64,13 @@ func (r *RoutingReconciler) ReconcileRouting(ctx context.Context, nebariApp *app
...
+		return err
+	}
```
dcmcand, line 73: **Question:** the gateway-not-found branch above (line 60) emits a Warning event via `r.Recorder.Event(...)` in addition to setting the condition, while this build-failure branch only sets the condition. Other failure paths in this file (CreationFailed, UpdateFailed) also skip the event, so precedent is mixed - but emitting one would mean operators see the failure in `kubectl describe nebariapp` output without having to read conditions. Optional.

internal/controller/reconcilers/routing/httproute.go
```diff
@@ -302,15 +310,21 @@ func (r *RoutingReconciler) ReconcilePublicRoute(ctx context.Context, nebariApp
 	logger.Info("Reconciling public route", "gateway", gatewayName, "hostname", nebariApp.Spec.Hostname,
 		"publicRoutes", nebariApp.Spec.Routing.PublicRoutes)
 
-	desiredRoute := r.buildPublicHTTPRoute(nebariApp, gatewayName, tlsListenerName)
+	desiredRoute, err := r.buildPublicHTTPRoute(nebariApp, gatewayName, tlsListenerName)
+	if err != nil {
+		logger.Error(err, "Failed to build public HTTPRoute")
+		conditions.SetCondition(nebariApp, appsv1.ConditionTypeRoutingReady, metav1.ConditionFalse,
+			"BuildFailed", fmt.Sprintf("Failed to build public HTTPRoute: %v", err))
+		return err
```
dcmcand, line 318: **Question (pre-existing, surfaced by this PR):** this new branch sets `RoutingReady=False`, but the create-error and update-error branches further down in `ReconcilePublicRoute` return the error without setting any condition. Result: a build failure is now observable on the resource, but a public-route create/update failure still isn't. Out of scope for this PR, but worth a follow-up issue to make the function symmetric.

internal/controller/reconcilers/routing/httproute_test.go
```diff
@@ -453,6 +511,60 @@ func TestReconcileRouting(t *testing.T) {
 	}
 }
 
+func TestReconcileRouting_BuildError(t *testing.T) {
+	// Use a properly initialised scheme for the fake client so validateGateway
+	// passes, but give the reconciler an empty scheme so SetControllerReference
+	// fails — confirming ReconcileRouting surfaces the error and sets the
+	// RoutingReady condition to False.
```
dcmcand, line 518: Clever - using a fully-populated scheme on the fake client so `validateGateway` passes, then giving the reconciler an empty scheme so `SetControllerReference` fails. Lets you exercise the build-error branch end-to-end without any mocking framework. The inline comment is a nice touch since the trick isn't obvious.

internal/controller/reconcilers/routing/httproute_test.go
```diff
@@ -256,6 +259,61 @@ func TestBuildHTTPRoute(t *testing.T) {
...
+		},
+	}
+
+	route, err := reconciler.buildHTTPRoute(nebariApp, "nebari-gateway", "")
+	if err == nil {
+		t.Error("expected error when scheme has no types registered, got nil")
+	}
+	if route != nil {
+		t.Error("expected nil route on error, got non-nil")
+	}
```
dcmcand, line 286: Good that this asserts `route != nil` (well, `route` is nil) in addition to the error - catches the case where a future refactor returns a partially-built route alongside an error.

dcmcand (approved):
Nice fix. The orphan-HTTPRoute risk from a swallowed `SetControllerReference` error is real, and the diff propagates the error cleanly through both `ReconcileRouting` and `ReconcilePublicRoute` with condition wiring on the failure path. Test coverage is solid: `TestReconcileRouting_BuildError` is a clever way to trigger the failure without mocks (split-scheme so `validateGateway` passes but `SetControllerReference` fails), and the builder-level tests assert `route == nil` on error in addition to checking the error itself.

Approving. A couple of non-blocking observations as inline comments, plus one general thought below about pre-existing asymmetry in `ReconcilePublicRoute` that's worth a follow-up.

## General notes

**Pre-existing condition-handling asymmetry in `ReconcilePublicRoute`.** This PR makes the build-error path observable on the resource, which is great, but the existing create-error and update-error branches in the same function (around the create at the not-found path and the update path further down) still return the error without setting `RoutingReady=False`. The commit message says "Set RoutingReady=False on both error paths so the failure is observable" - true for the build path, but the function as a whole is still uneven. Out of scope here; worth a follow-up issue.

**Naming the new reason.** `"BuildFailed"` is consistent with neighboring inline literals (`"CreationFailed"`, `"UpdateFailed"`), but the API package (`api/v1/nebariapp_types.go`) defines `EventReasonGatewayNotFound` and friends as constants. Adding `EventReasonHTTPRouteBuildFailed` and using it at both call sites would make the reason discoverable from the API surface. Cleaning up the other inline literals could be a separate PR.

**Documenting the rule.** It would be useful to add a short subsection to `docs/reconcilers/routing.md` codifying "builders that own children must propagate `SetControllerReference` errors and set `RoutingReady=False`; silently discarding (`_ = ...`) is forbidden." The convention exists already in `auth/reconciler.go` and now here, but it's only enforced by example.

**Table-driven tests.** Repo convention is table-driven Go tests. The two new builder error tests have distinct setup (different scheme wiring), so a table would be near-empty, but they could be folded into the existing `TestBuildHTTPRoute` / `TestBuildPublicHTTPRoute` tables with an `expectError bool` column. Preference, not a blocker.

PR #108, opened 2026-04-16
Fix token exchange decision strategy to AFFIRMATIVE

tylerpotts:
## Summary
- Changes the token-exchange scope permission `decisionStrategy` from `UNANIMOUS` to `AFFIRMATIVE`
- With `UNANIMOUS`, all peer client policies must match — impossible since each represents a different client. Any token exchange request fails with `403: Client not allowed to exchange`
- With `AFFIRMATIVE`, any single matching peer client policy grants access, which is the correct behavior

## Test plan
- [x] Deploy nebi-pack with `tokenExchange.enabled: true` alongside data-science-pack
- [x] Verify JupyterHub can exchange tokens for a nebi-audience token via Keycloak
- [x] Verify multiple peer clients (jupyterhub, mlflow, etc.) can each independently exchange tokens


viniciusdc (approved):
Very well explained context, thanks!

PR #109, opened 2026-04-21
feat(auth): expose forwardAccessToken on AuthConfig

dcmcand:
## Summary

Adds `AuthConfig.ForwardAccessToken` (`*bool`) to the `NebariApp` CRD. The operator passes it through to the generated `SecurityPolicy.spec.oidc.forwardAccessToken` field on the Envoy Gateway side.

When `true`, Envoy's OIDC filter forwards the OAuth2 access token to the upstream backend as `Authorization: Bearer <token>`. Without this, the gateway only stores tokens in encrypted session cookies (`IdToken-*`) which backends cannot decode.

## Why

Backends that need to read the JWT for per-user authorization decisions can't do so without this. Concrete blocker we hit: the `nebari-llm-serving` key-manager UI filters models by the user's `groups` claim. Because the operator-generated SP didn't include `forwardAccessToken`, the key-manager only ever received Envoy's encrypted session cookie, parsed no groups, and returned an empty model list to every user regardless of their actual group membership.

## Behaviour

- Unset (default): OIDC field left nil → Envoy default applies (no forwarding). No behaviour change for existing NebariApps.
- `true`: ` Authorization: Bearer <access_token>` appears on upstream requests.
- Only applies when `enforceAtGateway: true` (consistent with the rest of the gateway-enforced auth fields).

## Test plan

- [x] Unit tests for both states added to ` TestBuildSecurityPolicySpec`
- [x] ` go test ./...` clean
- [x] CRD regenerated; ` config/crd/bases/reconcilers.nebari.dev_nebariapps.yaml` updated
- [ ] Smoke test on a live cluster: deploy this operator image, set ` spec.auth.forwardAccessToken: true` on the key-manager NebariApp, verify the resulting SP has the field and the UI lists models for an authenticated user

## Draft because

- Pending live-cluster smoke test (in-flight on llmd-test)
- May want to add validation that ` forwardAccessToken: true` rejects when ` enforceAtGateway: false`


api/v1/nebariapp_types.go
```diff
@@ -216,6 +216,17 @@ type AuthConfig struct {
...
+	// ForwardAccessToken instructs the gateway-enforced OIDC filter to forward
+	// the user's OAuth2 access token to the upstream service via the
+	// `Authorization: Bearer <token>` header. Use this when the application
+	// needs to read the JWT itself - for example to extract the user's groups
+	// claim and apply per-user authorization decisions on top of the gateway's
+	// authentication. By default the gateway only stores the token in an
+	// encrypted session cookie that backends cannot decode.
+	// Only applies when enforceAtGateway is true.
+	// +optional
+	ForwardAccessToken *bool `json:"forwardAccessToken,omitempty"`
```
viniciusdc, line 229: Nothing prevents a user from setting `forwardAccessToken: true` with `enforceAtGateway: false`. In that case, the field is silently ignored — `buildSecurityPolicySpec` is never called when gateway enforcement is off, so no `SecurityPolicy` is created.

Since the field is brand-new, this is a good time to lock it down with a CEL validation rule:

```go
// +kubebuilder:validation:XValidation:rule="!has(self.forwardAccessToken) || (has(self.enforceAtGateway) && self.enforceAtGateway == true)",message="forwardAccessToken requires enforceAtGateway: true"
```

This fails with `kubectl apply` rather than leaving the user to debug why the backend never sees a token. The PR body already flags this under 'Draft because' — just noting it here for visibility.

internal/controller/reconcilers/auth/reconciler.go
```diff
@@ -465,6 +465,15 @@ func (r *AuthReconciler) buildSecurityPolicySpec(ctx context.Context, nebariApp
 		oidcConfig.Scopes = []string{"openid", "profile", "email"}
 	}
 
+	// Forward the OAuth2 access token to the upstream as Authorization: Bearer
+	// when the user opts in. Applications that read the JWT for per-user
+	// authorization (e.g. group claim filtering) need this; without it the
+	// gateway only stores the token in an encrypted session cookie that
+	// backends cannot decode.
+	if nebariApp.Spec.Auth.ForwardAccessToken != nil {
```
viniciusdc, line 473:
```go
oidcConfig.ForwardAccessToken = nebariApp.Spec.Auth.ForwardAccessToken
```

This copies the pointer, so both the `NebariApp` spec and the constructed `SecurityPolicySpec` now point to the same `bool`. If anything mutates the spec after construction it'll silently affect the upstream object too. Sibling fields use `ptrTo(...)` to allocate fresh copies (`ClientID`, `RedirectURL`, `LogoutPath`). Prefer:

```go
oidcConfig.ForwardAccessToken = ptrTo(*nebariApp.Spec.Auth.ForwardAccessToken)
```

internal/controller/reconcilers/auth/reconciler.go
```diff
@@ -465,6 +465,15 @@ func (r *AuthReconciler) buildSecurityPolicySpec(ctx context.Context, nebariApp
 		oidcConfig.Scopes = []string{"openid", "profile", "email"}
 	}
 
+	// Forward the OAuth2 access token to the upstream as Authorization: Bearer
+	// when the user opts in. Applications that read the JWT for per-user
+	// authorization (e.g. group claim filtering) need this; without it the
+	// gateway only stores the token in an encrypted session cookie that
+	// backends cannot decode.
```
viniciusdc, line 472: `forwardAccessToken: true` materially changes what downstream services receive. A single debug line makes this visible when operators are tracing why a backend suddenly started or stopped seeing tokens:

```go
logger := log.FromContext(ctx)
logger.V(1).Info("forwarding access token to upstream as Authorization: Bearer")
```

The other gateway switches (DenyRedirect, scopes) are also silent, so you're consistent with the current style — flagging this because the token-forwarding behavior is harder to debug without a trace.

internal/controller/reconcilers/auth/reconciler_test.go
```diff
@@ -555,6 +555,84 @@ func TestBuildSecurityPolicySpec(t *testing.T) {
 	}
 }
 
+// TestBuildSecurityPolicySpec_ForwardAccessToken covers the forwardAccessToken
+// passthrough in isolation. Kept separate from TestBuildSecurityPolicySpec
+// to keep that table-driven test below the gocyclo complexity threshold.
+func TestBuildSecurityPolicySpec_ForwardAccessToken(t *testing.T) {
+	scheme := runtime.NewScheme()
```
viniciusdc, line 564: Three states map naturally to one pointer: `nil`, `ptrTo(true)`, `ptrTo(false)`. The assertion then becomes:

```go
if !reflect.DeepEqual(spec.OIDC.ForwardAccessToken, tt.want) {
    t.Errorf("ForwardAccessToken = %v, want %v", spec.OIDC.ForwardAccessToken, tt.want)
}
```

Harder to mis-read than the two-bool combination. Not blocking — current approach is correct.

viniciusdc (commented):
Some small nitpicks

api/v1/nebariapp_types.go
viniciusdc: We could leave as is, and silently ignore the flag since it won't apply, but I can see a case where a user is trying out the config options and asking themselves why its not working ksksks 😆

viniciusdc (changes requested):

viniciusdc:
Quick question on the CEL rule
https://github.com/nebari-dev/nebari-operator/blob/39421a81bf480308572a7315b0ecd85312ff66a0/api/v1/nebariapp_types.go#L157

The current rule rejects any *presence* of `forwardAccessToken` when `enforceAtGateway: false`, including `forwardAccessToken: false`. So a user who explicitly opts out of both ends up getting an error that says "forwardAccessToken requires enforceAtGateway: true", which is a bit confusing.

Also, having a test in case that line is ever removed might be good too

viniciusdc (approved):

viniciusdc (approved):
Wonderful, thanks @dcmcand !

PR #114, opened 2026-05-06
feat: support user-provided TLS secrets via routing.tls.secretName (#90)

### LINKED ISSUE #90: Support custom/pre-provisioned TLS certificates via existingSecret reference
## Problem

Currently, the operator always creates cert-manager Certificate resources that obtain certificates through a configured ClusterIssuer. While users can work around this by using a CA-based ClusterIssuer with a pre-provisioned CA certificate, there's no way to directly reference an existing TLS secret for a NebariApp.

This limitation affects environments that:
- Use pre-provisioned certificates from external sources
- Cannot run cert-manager at all
- Need to use certificates provisioned outside the cluster
- Have certificates managed by external tooling/processes
- Require certificate rotation to happen outside Kubernetes

## Proposed Solution

Add an optional field to `RoutingTLSConfig` that allows referencing an existing Kubernetes TLS secret instead of creating a cert-manager Certificate:

```yaml
apiVersion: reconcilers.nebari.dev/v1
kind: NebariApp
metadata:
  name: myapp
spec:
  hostname: myapp.example.com
  service:
    name: myapp-service
    port: 8080
  routing:
    tls:
      enabled: true
      existingSecret: my-custom-tls-cert  # New optional field
```

## Implementation Considerations

1. **Behavior when `existingSecret` is set:**
   - Skip cert-manager Certificate creation
   - Still create the per-app Gateway HTTPS listener
   - Reference the specified secret in the Gateway listener's `certificateRefs`
   - Validate that the secret exists and is of type `kubernetes.io/tls`

2. **Secret namespace:**
   - The secret must be in the Gateway namespace (`envoy-gateway-system`)
   - Document requirement for users to create/maintain the secret in correct namespace
   - Consider adding validation to check secret existence

3. **Mutual exclusivity:**
   - `existingSecret` and ClusterIssuer-based certificate creation are mutually exclusive
   - When `existingSecret` is set, ignore `TLS_CLUSTER_ISSUER_NAME`

4. **Status conditions:**
   - Adjust `TLSReady` condition logic to check secret existence instead of Certificate readiness
   - Consider adding events/warnings if secret is missing or invalid

## Benefits

- Simplifies deployment in environments without ACME access
- Supports certificate management workflows external to the cluster
- Maintains backward compatibility (existing behavior when omitted)
- Still leverages Gateway API for routing (only certificate source changes)

## Related Discussion

This enhancement was discussed in the context of air-gapped and restricted environments where DNS configuration and ACME providers are not accessible.

dcmcand:
Closes #90.

## Summary

Adds an optional `routing.tls.secretName` field to `NebariApp` so an app can reference a pre-existing Kubernetes TLS secret instead of forcing cert-manager to issue a per-app certificate. Useful for air-gapped environments, externally managed certs (wildcards, corporate CAs), or any case where running cert-manager isn't viable.

When `secretName` is set, the operator:

- skips cert-manager `Certificate` creation and cleans up any previously owned `Certificate` (supports migration from the cert-manager path)
- points the per-app Gateway HTTPS listener at the user's secret in `envoy-gateway-system`
- best-effort-checks the secret on every reconcile and reports `TLSReady` with reasons `UserProvidedSecretReady` / `UserProvidedSecretNotFound` / `UserProvidedSecretInvalidType` (a missing secret does not block reconciliation - the listener is still attached so Envoy picks it up as soon as the user creates it)
- emits matching events: `UserProvidedSecretInUse`, `UserProvidedSecretNotFound`, `UserProvidedSecretInvalid`

When `secretName` is unset, behavior is unchanged.

Also: the TLS reconciler is now always wired up in `cmd/operator/main.go`. Previously it was only constructed when `TLS_CLUSTER_ISSUER_NAME` was set, which silently meant a user-provided-secret NebariApp had no reconciler running on a cluster without a ClusterIssuer. The reconciler itself already reports `ClusterIssuerNotConfigured` on the cert-manager path when no issuer is set, so always wiring it up is safe.

## Changes

- `api/v1`: new `SecretName` field on `RoutingTLSConfig`; new condition and event reason constants
- `internal/controller/reconcilers/tls`: branch in `ReconcileTLS`; new helpers `cleanupOwnedCertificate` and `checkUserProvidedSecret`; `reconcileGatewayListener` parameterized by secret name
- `cmd/operator/main.go`: always initialize the TLS reconciler
- `config/crd`, `api/v1/zz_generated.deepcopy.go`, `docs/api-reference.md`: regenerated
- `docs/configuration-reference.md`, `docs/reconcilers/routing.md`: documents the new field with example and migration notes
- `test/e2e/tls_user_secret_test.go`: e2e spec verifying TLSReady reason, no Certificate created, listener references the user secret

## Example

```yaml
apiVersion: reconcilers.nebari.dev/v1
kind: NebariApp
metadata:
  name: myapp
  namespace: myteam
spec:
  hostname: myapp.example.com
  service:
    name: myapp-backend
    port: 8080
  routing:
    tls:
      enabled: true
      secretName: myteam-wildcard-tls
```

The secret must live in `envoy-gateway-system` and be of type `kubernetes.io/tls`.

## Test plan

- [x] Unit tests pass: `go test ./internal/controller/reconcilers/... ./internal/controller/utils/... ./internal/config/...`
- [x] Lint clean: `make lint`
- [x] Build clean, no manifest/deepcopy drift: `make manifests generate build`
- [x] E2E test passes against Kind: `USE_EXISTING_CLUSTER=true go test ./test/e2e -tags=e2e -ginkgo.focus="routing.tls.secretName"`
- [x] CI green
- [ ] Manual: confirm reason transitions on a real cluster as the secret is created/deleted/changed type


internal/controller/reconcilers/tls/reconciler.go
```diff
@@ -142,14 +159,63 @@ func (r *TLSReconciler) ReconcileTLS(ctx context.Context, nebariApp *appsv1.Neba
...
+			conditions.SetCondition(nebariApp, appsv1.ConditionTypeTLSReady, metav1.ConditionFalse,
+				"GatewayListenerFailed", fmt.Sprintf("Failed to reconcile Gateway listener: %v", err))
+		}
+		return nil, err
+	}
+
+	status, reason, msg := r.checkUserProvidedSecret(ctx, secretName)
+	conditions.SetCondition(nebariApp, appsv1.ConditionTypeTLSReady, status, reason, msg)
+
+	switch reason {
```
viniciusdc: These events fire on every reconcile pass — every ~30s when the secret is missing, every ~1min when healthy. Kubernetes deduplicates identical `reason`+`message` pairs (increments `Count`), so it won't flood the API server, but `kubectl describe` becomes noisy over time and you lose the signal of "when did this transition happen."

Consider only emitting when the reason changes from the previous condition:

```go
prev := conditions.GetCondition(nebariApp, appsv1.ConditionTypeTLSReady)
if prev == nil || prev.Reason != reason {
    switch reason {
    case appsv1.ReasonUserProvidedSecretReady:
        r.Recorder.Event(...)
    // ...
    }
}
```

internal/controller/reconcilers/tls/reconciler.go
```diff
@@ -434,6 +499,79 @@ func (r *TLSReconciler) removeGatewayListener(ctx context.Context, nebariApp *ap
...
+				appsv1.ReasonUserProvidedSecretNotFound,
+				fmt.Sprintf("TLS secret %s/%s not found; create it and the listener will pick it up",
+					constants.GatewayNamespace, secretName)
+		}
+		return metav1.ConditionFalse,
+			appsv1.ReasonUserProvidedSecretCheckFailed,
+			fmt.Sprintf("failed to check TLS secret %s/%s: %v", constants.GatewayNamespace, secretName, err)
+	}
+
+	if secret.Type != corev1.SecretTypeTLS {
```
viniciusdc, line 577: The type check is necessary but not quite sufficient. A `kubernetes.io/tls` secret can be created via direct API (e.g. `kubectl apply -f` with a raw YAML manifest) with empty or missing `tls.crt`/`tls.key` data — `kubectl create secret tls` validates and requires real PEM files, but nothing stops a user from creating the secret without going through that command. Envoy would silently fail to serve TLS while the operator reports `UserProvidedSecretReady`.

Worth adding a data-key check after this block:

```go
if len(secret.Data["tls.crt"]) == 0 || len(secret.Data["tls.key"]) == 0 {
    return metav1.ConditionFalse,
        appsv1.ReasonUserProvidedSecretInvalidType,
        fmt.Sprintf("TLS secret %s/%s has type kubernetes.io/tls but is missing tls.crt or tls.key data",
            constants.GatewayNamespace, secretName)
}
```

Could reuse `ReasonUserProvidedSecretInvalidType` to keep the reason space small, or add `ReasonUserProvidedSecretMissingData` if you want alerting rules to distinguish "wrong type" from "right type, empty data".

internal/controller/reconcilers/tls/reconciler.go
```diff
@@ -142,14 +159,63 @@ func (r *TLSReconciler) ReconcileTLS(ctx context.Context, nebariApp *appsv1.Neba
...
+	}
+
+	return &TLSResult{
+		ListenerName: naming.ListenerName(nebariApp),
+		SecretName:   secretName,
+		CertReady:    status == metav1.ConditionTrue,
+	}, nil
+}
+
 // CleanupTLS removes TLS resources for a NebariApp.
```
viniciusdc, line 231: `CleanupTLS` calls `deleteCertificate` (line 233), which deletes by name only — no ownership-label check. The live reconcile path uses `cleanupOwnedCertificate`, which verifies `nebari.dev/nebariapp-name` / `nebari.dev/nebariapp-namespace` labels before deleting. The scenario where an unowned cert shares the same derived name is unlikely, but the two paths are inconsistent and `deleteCertificate` is the riskier one to have on the deletion path.

One backward-compatible option: add an optional `OwnerLabels map[string]string` field to `TLSReconciler`. When non-nil, `CleanupTLS` does label-aware deletion; when nil it keeps the current unconditional behavior. Setting it to nil is the default so nothing breaks for existing deployments.

Simpler alternative: since `cleanupOwnedCertificate` is already idempotent and handles not-found gracefully, `CleanupTLS` could just call it in place of `deleteCertificate` on both paths — it is strictly safer and costs nothing extra.

viniciusdc (commented):
**`internal/controller/nebariapp_controller.go:148–151` — stale comment (not in diff, flagging here)**

The block above `if r.TLSReconciler != nil` still reads:

> "When TLSReconciler is nil (TLS_CLUSTER_ISSUER_NAME not set), no per-app Certificate or Gateway listener is created."

With this PR `main.go` always constructs a non-nil `TLSReconciler`, so this comment describes the old world. The nil guard at line 153 is still fine for tests, but the comment should be updated to reflect that the reconciler is always wired up and the `ClusterIssuerNotConfigured` condition handles the no-issuer case internally.

viniciusdc (approved):

oren-openteams:
Re-tested against `quay.io/nebari/nebari-operator@sha256:8223484e…` with ArgoCD auto-sync paused this time. Both fixes verified:

- ✅ `UserProvidedSecretInUse` event fires on transition into Ready. Operator log at `2026-05-20T16:53:11Z` shows the Recorder call landing with `reason: UserProvidedSecretInUse` after I flipped `secretName` back to a real secret. (My earlier "still silent" report was wrong — I was using `kubectl get events --field-selector reason=X > file`, and the empty stdout when no events match looked like the event was missing.)
- ✅ TLSReady condition cycled `UserProvidedSecretReady → UserProvidedSecretNotFound → UserProvidedSecretReady` cleanly. No stuck condition.

Good to merge from my side.

oren-openteams (approved):
Verified via the test described in the comment above — both fixes work as intended.

Issue #139, opened 2026-06-06
Epic: Routing & SecurityPolicy consolidation


viniciusdc:
## Goal

Consolidate the cluster of open issues that all land on the operator's
**HTTPRoute** and **SecurityPolicy** generation into a sequenced mix of one bug fix
and additive work. Several of these issues touch the same two reconcilers, and the
duplicated route/endpoint-construction code is the common root: unifying it first
turns most of the remaining fixes into small additive diffs.

This is a tracking issue — it does not replace the linked issues, it orders them.
The definition of done for each item lives in its linked issue. The numbering is a
priority ordering, not a dependency chain — PR 1 goes first because it fixes an
active 500, but the three PRs touch different code and can land in any order.

## Why these belong together

Everything below lands on one of three surfaces:

- **HTTPRoute construction** — `internal/controller/reconcilers/routing/httproute.go`.
  `buildHTTPRoute`/`buildHTTPRouteRules` and `buildPublicHTTPRoute` (with its inline
  rule) are near-duplicates: same parentRef / TLS-section / label / backend logic,
  written twice. Any change that has to land "on both routes" (filters, overlap
  detection) currently means editing two places.
- **SecurityPolicy / OIDC endpoint derivation** — `auth/reconciler.go` +
  `providers/keycloak.go`. The issuer/endpoint URLs are assembled independently in
  `buildIssuerURL`, `GetIssuerURL`, `GetEndpointOverrides`, and
  `GetExternalIssuerURL`. The bugs below are all "we resolved the wrong URL for
  endpoint X."
- **Controller wiring** — `nebariapp_controller.go` `SetupWithManager`. The
  operator only watches `NebariApp` via `For()`; owned resources reconcile only on
  the periodic requeue.

### PR 1 — OIDC endpoint resolution (bug, fast-track)

Release-relevant: enabling auth on the operator's own quickstart currently returns
**HTTP 500** because the generated SecurityPolicy points at an unreachable Keycloak
URL. Decouple from the additive work below so the fix can ship on its own.

- [ ] #136 — issuer built as `:8080` with no `/auth` context path → OIDC discovery
  times out. Root cause is the `IssuerServicePort` (default `8080`) /
  `IssuerContextPath` (default `""`) defaults combined with `manager.yaml` not
  setting `KEYCLOAK_ISSUER_SERVICE_PORT` / `KEYCLOAK_ISSUER_CONTEXT_PATH`. Fix spans
[...]

Issue #113, opened 2026-04-27
Add reconciler test asserting SecurityPolicy.spec.oidc.provider has Token on in-cluster URL and Authorization on public URL


dcmcand:
## Context

PR #111 added unit tests on \`KeycloakProvider.GetEndpointOverrides\` covering the struct values it returns. The reconciler-level test (\`TestBuildSecurityPolicySpec\` in \`internal/controller/reconcilers/auth/reconciler_test.go\`) does not currently exercise the split-endpoint behavior end-to-end.

## Ask

Add a reconciler test that builds a \`SecurityPolicy\` via \`buildSecurityPolicySpec\` with a \`KeycloakProvider\` configured with both internal service config and \`KEYCLOAK_EXTERNAL_URL\`, and asserts:

- \`spec.OIDC.Provider.TokenEndpoint\` host equals the in-cluster service hostname
- \`spec.OIDC.Provider.AuthorizationEndpoint\` host equals the \`ExternalURL\` host
- \`spec.OIDC.Provider.EndSessionEndpoint\` host equals the \`ExternalURL\` host

Catches future regressions where the override-application logic in \`reconciler.go\` gets refactored without touching \`keycloak_test.go\`.

## Reference

Stylistic note from code review of #111.

Issue #112, opened 2026-04-27
OIDC SecurityPolicy.Issuer field stays on in-cluster URL; investigate iss-claim mismatch with Keycloak frontendUrl deployments


dcmcand:
## Context

PR #111 split the SecurityPolicy OIDC endpoint overrides into browser-facing (Authorization, EndSession - public URL via `KEYCLOAK_EXTERNAL_URL`) and back-channel (Token - in-cluster URL). The Issuer field was deliberately not changed: `GetIssuerURL` still returns the in-cluster realm URL.

## Concern

When Keycloak is configured with a public `frontendUrl` (recommended for production), it issues tokens with the `iss` claim set to the public URL. Envoy's `SecurityPolicy.spec.oidc.provider.issuer` is the in-cluster URL. Whether this is a problem depends on whether Envoy Gateway's OAuth2 filter validates the `iss` claim against the configured Issuer field. The OAuth2 filter mainly uses Issuer for OIDC discovery, but downstream consumers (e.g., LLM serving pack JWT validation) may compare the iss claim against an expected issuer string.

## Investigate

1. Does Envoy Gateway's OAuth2 filter validate \`iss\` against \`provider.issuer\` beyond using it for discovery? If yes, a mismatch will fail token validation.
2. Do downstream services that read the \`id_token\` cookie compare \`iss\` against an expected value?
3. Should \`provider.Issuer\` be set from \`KEYCLOAK_EXTERNAL_URL\` (matching the iss claim Keycloak actually emits), with discovery falling through to the externally-fetched well-known doc?

## Acceptance

A clear answer to whether the Issuer field needs to track ExternalURL too, with a test case demonstrating the chosen behavior.

## Reference

Surfaced during code review of #111. Validation of the Nebari LLM Serving Pack fresh-install runbook (`nebari-dev/nebari-llm-serving-pack#65`) is the originating context.

PR #140, opened 2026-06-06
fix: default Keycloak issuer to keycloakx layout (port 80, /auth)

### LINKED ISSUE #136: Generated SecurityPolicy uses wrong Keycloak port (8080) and missing /auth path prefix
## Summary

Enabling auth on a `NebariApp` doesn't redirect users to the Keycloak login page; requests return `HTTP 500` instead. The operator generates a `SecurityPolicy` with an unreachable Keycloak URL, so Envoy Gateway can't enforce the OIDC flow.

## Steps to reproduce

These steps follow the operator's own [quickstart](https://github.com/nebari-dev/nebari-operator/blob/main/docs/quickstart.md), which is what surfaced the bug.

1. Fresh dev cluster: `cd dev && make setup`
2. Install operator: `make docker-build IMG=... && kind load ... && make install && make deploy IMG=...`
3. Set up the Keycloak realm: `./dev/scripts/services/keycloak/setup.sh`
4. Create a namespace, deploy a sample Deployment + Service, then apply a `NebariApp` with auth enabled:

```bash
kubectl create namespace demo
kubectl label namespace demo nebari.dev/managed=true

kubectl apply -f - <<EOF
apiVersion: apps/v1
kind: Deployment
metadata:
  name: hello-app
  namespace: demo
spec:
  replicas: 1
  selector:
    matchLabels:
      app: hello
  template:
    metadata:
      labels:
        app: hello
    spec:
      containers:
      - name: hello
        image: hashicorp/http-echo
        args: ["-text=Hello"]
        ports:
        - containerPort: 5678
---
apiVersion: v1
kind: Service
metadata:
  name: hello-service
  namespace: demo
spec:
  selector:
    app: hello
  ports:
  - port: 80
    targetPort: 5678
---
apiVersion: reconcilers.nebari.dev/v1
kind: NebariApp
metadata:
  name: hello-app
  namespace: demo
spec:
  hostname: hello-app.nebari.local
  service:
    name: hello-service
    port: 80
  routing:
    tls:
      enabled: true
  auth:
    enabled: true
    provider: keycloak
    provisionClient: true
    scopes:
      - openid
      - profile
      - email
EOF
```

5. Inspect the operator-generated `SecurityPolicy`:

```bash
kubectl describe securitypolicy hello-app-security -n demo
```

## Expected

In the `Status` block at the bottom of the describe output, the `Accepted` condition should look like:

```
Status:
  Ancestors:
    Conditions:
      Status:  True
      Type:    Accepted
```

And requests to the app's hostname should redirect to the Keycloak login page (`HTTP 302`).

## Actual

In the same `Status` block, the `Accepted` condition is stuck on `False` with an OIDC error:

```
Status:
  Ancestors:
    Conditions:
      Message:  OIDC: error fetching endpoints from issuer:
                  Get "http://keycloak-keycloakx-http.keycloak.svc.cluster.local:8080/realms/nebari/.well-known/openid-configuration":
                  dial tcp 10.96.71.87:8080: i/o timeout.
      Reason:   Invalid
      Status:   False
      Type:     Accepted
```

And requests to the app's hostname return `HTTP 500` instead of redirecting:

```bash
curl -k -I https://hello-app.nebari.local
```

Output:

```
HTTP/2 500
date: ...
```

## Root cause

The keycloakx Helm chart (the one `dev/services-install` uses) exposes Keycloak at:

- **Service port**: `80` (not `8080`)
- **Context path**: `/auth` (not the root)

`kubectl get svc -n keycloak`:

```
NAME                      TYPE        PORT(S)
keycloak-keycloakx-http   ClusterIP   9000/TCP,80/TCP,8443/TCP
```

A quick reachability test from inside the cluster (e.g., `kubectl run -i --rm test-curl --image=curlimages/curl -- curl -s -o /dev/null -w "%{http_code}\n" <URL>`):

| URL | Result |
| --- | --- |
| `http://keycloak-keycloakx-http.keycloak.svc.cluster.local:8080/realms/nebari/.well-known/openid-configuration` | timeout (no listener on 8080) |
| `http://keycloak-keycloakx-http.keycloak.svc.cluster.local:80/realms/nebari/.well-known/openid-configuration` | `404` |
| `http://keycloak-keycloakx-http.keycloak.svc.cluster.local:80/auth/realms/nebari/.well-known/openid-configuration` | `200` ✓ |

The operator generates:

```
http://keycloak-keycloakx-http.keycloak.svc.cluster.local:8080/realms/nebari
```

The correct issuer for this dev install would be:

```
http://keycloak-keycloakx-http.keycloak.svc.cluster.local/auth/realms/nebari
```

viniciusdc:
## Summary

Enabling auth on a `NebariApp` returned **HTTP 500** instead of redirecting to Keycloak, reproducible straight from the operator's own quickstart. The generated `SecurityPolicy` pointed Envoy at an unreachable Keycloak issuer URL, so OIDC discovery timed out.

This is PR 1 of the routing/SecurityPolicy consolidation tracked in #139.

## Root cause

The OIDC issuer is built from `DefaultKeycloakServicePort` (`8080`) and `DefaultKeycloakContextPath` (`""`). But `DefaultKeycloakServiceName` already hardcodes the `keycloak-keycloakx-http` service, and the codecentric/keycloakx chart serves HTTP on **port 80 under `/auth`** (its `http.relativePath`, which `dev/scripts` also set explicitly). The defaults were internally inconsistent: they named the keycloakx service but used a port/path that service never listens on.

Result — the operator generated:

```
http://keycloak-keycloakx-http.keycloak.svc.cluster.local:8080/realms/nebari
```

Envoy dialed `:8080` (no listener) → `dial tcp ...:8080: i/o timeout` → `SecurityPolicy` stuck `Accepted=False` → 500.

## Change

- **`constants.go`** — `DefaultKeycloakServicePort` `8080 → 80`, `DefaultKeycloakContextPath` `"" → "/auth"`, so the issuer is reachable out of the box on the chart the operator already targets. Both remain overridable via `KEYCLOAK_ISSUER_SERVICE_PORT` / `KEYCLOAK_ISSUER_CONTEXT_PATH` for a root-path Keycloak.
- **`config/manager/manager.yaml`** — set those two env vars explicitly so the deployed issuer is obvious in the manifest and easy to retarget.
- **`reconciler_test.go`** — new `TestBuildSecurityPolicySpec_KeycloakSplitEndpoints` exercising the split-endpoint behavior end to end with a real `KeycloakProvider`: Token stays in-cluster, Authorization/EndSession use the public `ExternalURL` host. Full-URL assertions also guard the issuer fix here.
- **`auth_test.go`** — updated the default-derived expectations; the explicit-override case (`9090` / explicit `""`) is unchanged.

## Test plan

- `go build ./...` and `go vet ./...` clean.
- `go test ./internal/config/... ./internal/controller/reconcilers/auth/...` — pass, including the new split-endpoint test.
- The `internal/controller` envtest suite (`TestControllers`) fails locally on a scheme-registration gap that also fails on clean `main` (unrelated to this change; needs `make test`'s CRD setup).

## Scope

- **Closes #136** (issuer port/path bug) and **#113** (split-endpoint reconciler test).
- **Does not** address #112 (whether `provider.issuer` must track `ExternalURL` under a Keycloak `frontendUrl`) — that's an open investigation, deferred per #139. The new test pins current behavior (issuer stays in-cluster) so #112's eventual change is a deliberate, visible diff.


viniciusdc:
need review, and extra work

PR #142, opened 2026-06-06
feat: watch owned resources and user-provided TLS secrets for prompt reconciliation

### LINKED ISSUE #30: enhancement: Add watches on owned resources (HTTPRoute, SecurityPolicy, Secret)
## Summary

The NebariApp controller only watches NebariApp resources via `For()` but does not watch the child resources it creates (HTTPRoute, SecurityPolicy, Secret). This means if an owned resource is externally modified or deleted, the controller won't be notified until the next periodic requeue (up to 1 minute).

## Problem

In `internal/controller/nebariapp_controller.go`:
```go
// nebariapp_controller.go:242-245
func (r *NebariAppReconciler) SetupWithManager(mgr ctrl.Manager) error {
    r.Recorder = mgr.GetEventRecorderFor("nebariapp-controller")
    return ctrl.NewControllerManagedBy(mgr).
        For(&appsv1.NebariApp{}).
        Named("nebariapp").
        Complete(r)
}
```

Missing watches for:
- `gatewayv1.HTTPRoute` - created by the routing reconciler with owner references
- `egv1alpha1.SecurityPolicy` - created by the auth reconciler with owner references
- `corev1.Secret` - created by the Keycloak provider for client credentials (note: no owner reference currently set on this secret)

### Impact
- If an HTTPRoute is manually deleted, traffic routing breaks until the next 1-minute requeue
- If a SecurityPolicy is modified externally, authentication configuration drifts
- The controller has no way to reconcile drift promptly

## Expected Behavior

The controller should use `.Owns()` to watch child resources, which triggers reconciliation when owned resources are created, updated, or deleted:

```go
return ctrl.NewControllerManagedBy(mgr).
    For(&appsv1.NebariApp{}).
    Owns(&gatewayv1.HTTPRoute{}).
    Owns(&egv1alpha1.SecurityPolicy{}).
    Owns(&corev1.Secret{}).
    Named("nebariapp").
    Complete(r)
```

Note: The OIDC client secret (`storeClientSecret` in `keycloak.go`) does not currently set owner references, so it would also need to be updated to set them for the `.Owns()` watch to work for secrets.

## Files

- `internal/controller/nebariapp_controller.go` - `SetupWithManager()` method (line 238)
- `internal/controller/reconcilers/auth/providers/keycloak.go` - `storeClientSecret()` method (line 294) - needs owner reference

## Definition of Done

- [ ] `SetupWithManager` uses `.Owns()` for HTTPRoute, SecurityPolicy, and Secret
- [ ] `storeClientSecret` sets owner reference on the OIDC client secret
- [ ] Gateway API and Envoy Gateway API types are registered in the scheme (verify existing scheme registration is sufficient)
- [ ] Integration test confirms that deleting an owned HTTPRoute triggers re-reconciliation and re-creation

### LINKED ISSUE #115: Watch user-provided TLS secrets to fast-track TLSReady transitions
## Background

PR #114 added \`routing.tls.secretName\` so a NebariApp can reference a pre-existing TLS secret in \`envoy-gateway-system\` instead of having cert-manager issue one. The reconciler is best-effort: when the secret is missing or has the wrong type, it still attaches the Gateway listener and reports \`TLSReady=False\` with reason \`UserProvidedSecretNotFound\` / \`UserProvidedSecretInvalidType\`.

## Problem

The controller does not watch \`Secret\` resources, so when a user creates the missing secret (or fixes the type), \`TLSReady\` only flips to \`True\` on the next periodic requeue. With the existing 30-second requeue when \`CertReady=false\` (\`internal/controller/nebariapp_controller.go:175\`), worst case is ~30 seconds. Acceptable for most setups, but it can show up as flaky readiness in CI/CD pipelines that deploy a NebariApp, then create the secret, then immediately curl the endpoint.

## Proposal

Add a filtered \`Watches\` on \`corev1.Secret\` in \`SetupWithManager\`:

\`\`\`go
builder.Watches(
    &corev1.Secret{},
    handler.EnqueueRequestsFromMapFunc(r.secretToNebariApp),
    builder.WithPredicates(predicate.NewPredicateFuncs(func(obj client.Object) bool {
        return obj.GetNamespace() == constants.GatewayNamespace
    })),
)
\`\`\`

\`secretToNebariApp\` lists NebariApps cluster-wide and returns reconcile requests for any whose \`spec.routing.tls.secretName\` equals the changed secret's name. Listing comes from the controller cache, so the cost per Secret event is negligible.

## Out of scope

- Watching cert-manager-issued secrets - those are already covered by the existing Certificate watch.
- Watching Secrets in app namespaces - the user-provided secret must live in \`envoy-gateway-system\` (the Gateway's namespace), so we only need to watch one namespace.

## Acceptance criteria

- Creating or modifying a Secret in \`envoy-gateway-system\` whose name matches the \`secretName\` of any NebariApp triggers reconciliation of that NebariApp within ~1 second.
- Unit test covering the map function (correct lookup, namespace filter, no spurious requeues for unrelated secrets).
- E2E test: deploy NebariApp with \`secretName\` pointing at a not-yet-existing secret, confirm \`TLSReady=False/UserProvidedSecretNotFound\`, create the secret, confirm \`TLSReady=True/UserProvidedSecretReady\` within a few seconds (well under the 30-second periodic requeue).

## Priority

Low. Behavior is correct without this, just slower than ideal.

## Related

- Implemented in #114 (initial user-provided TLS secret support)
- Issue #90 (parent feature request)

viniciusdc:
## Summary

PR 3 of the routing/SecurityPolicy consolidation (#139). The controller only watched `NebariApp` (via `For`) plus cert-manager `Certificate`s, so drift in the resources it manages was only repaired on the periodic requeue (up to ~1 minute). This adds event-driven watches.

## Owned resources (closes #30)

- `SetupWithManager` now `Owns` `HTTPRoute`, `SecurityPolicy`, and `Secret`, so external edits or deletions of those trigger reconciliation immediately.
- `HTTPRoute`s and `SecurityPolicy`s already carry controller owner references; the OIDC client `Secret` did not. `storeClientSecret` now sets one via `controllerutil.SetControllerReference` (and adopts pre-existing secrets on the next sync by writing the owner reference on update too).
- `KeycloakProvider` gains a `Scheme` field, wired from `mgr.GetScheme()` in `main.go`. A nil `Scheme` skips the owner reference, so existing provider tests that don't wire a scheme keep passing.
- Scheme registration already covers all three types (`gatewayapiv1.Install`, `egv1alpha1.AddToScheme`, core), so no scheme changes were needed.

## User-provided TLS secrets (closes #115)

- Adds a filtered `Watches` on `Secret`s in the Gateway namespace (`envoy-gateway-system`), mapped back to any NebariApp referencing them via `routing.tls.secretName`. Creating or fixing a referenced secret now flips `TLSReady` promptly instead of on the periodic requeue.
- The map function lists NebariApps from the controller cache, so the per-event cost is negligible. The namespace filter lives in the watch predicate; the OIDC client Secret in app namespaces is handled by `Owns` above, this watch is for the unowned, user-managed secret.

## Test plan

- `go build ./...`, `go vet ./...`, `gofmt -l` clean.
- `go test ./internal/controller/reconcilers/auth/...` passes, including the new `TestKeycloakProvider_StoreClientSecret_SetsOwnerReference`.
- `go test -run TestSecretToNebariApp ./internal/controller/` passes (placed in a standalone test file so it runs without the envtest suite).
- The `internal/controller` Ginkgo/envtest suite fails locally on a pre-existing scheme-registration gap (also fails on clean `main`; needs `make test`'s CRD setup) — unrelated to this change.

## Notes / follow-ups

- `Owns(&corev1.Secret{})` makes the manager cache Secrets in watched namespaces. That is what #30 asks for; if the cluster-wide Secret cache becomes a concern, a label-selector cache is a reasonable later optimization.
- #30's "integration test confirms deleting an owned HTTPRoute triggers re-creation" belongs in the envtest suite, which doesn't run in this environment. Left as a follow-up; the `Owns` wiring is the standard controller-runtime pattern and the owner-reference unit test covers the piece that was missing (the Secret owner ref).

### DIFF
--- a/internal/controller/nebariapp_controller.go
+++ b/internal/controller/nebariapp_controller.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	certmanagerv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
+	egv1alpha1 "github.com/envoyproxy/gateway/api/v1alpha1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
@@ -30,11 +31,14 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
 
 	appsv1 "github.com/nebari-dev/nebari-operator/api/v1"
 
@@ -382,22 +386,70 @@ func (r *NebariAppReconciler) cleanup(ctx context.Context, nebariApp *appsv1.Neb
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *NebariAppReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	builder := ctrl.NewControllerManagedBy(mgr).
+	b := ctrl.NewControllerManagedBy(mgr).
 		For(&appsv1.NebariApp{}).
-		Named("nebariapp")
+		Named("nebariapp").
+		// Watch the resources the operator owns so that external edits or
+		// deletions trigger prompt reconciliation (re-creation / drift repair)
+		// instead of waiting up to the periodic requeue interval. Owned objects
+		// are mapped back to their NebariApp via controller owner references:
+		// HTTPRoutes and SecurityPolicies set them at construction; the OIDC
+		// client Secret now sets one in storeClientSecret.
+		Owns(&gatewayv1.HTTPRoute{}).
+		Owns(&egv1alpha1.SecurityPolicy{}).
+		Owns(&corev1.Secret{})
 
 	// Watch cert-manager Certificates so that Certificate readiness transitions
 	// trigger NebariApp reconciliation without waiting for the periodic requeue.
 	// Certificates are matched to NebariApps via the nebari.dev/nebariapp-name
 	// and nebari.dev/nebariapp-namespace labels.
 	if r.TLSReconciler != nil {
-		builder = builder.Watches(
+		b = b.Watches(
 			&certmanagerv1.Certificate{},
 			handler.EnqueueRequestsFromMapFunc(r.certificateToNebariApp),
 		)
 	}
 
-	return builder.Complete(r)
+	// Watch user-provided TLS secrets in the Gateway namespace. When a NebariApp
+	// references routing.tls.secretName for a secret that does not yet exist (or
+	// has the wrong type), TLSReady stays False; creating or fixing the secret
+	// should flip it promptly rather than on the next periodic requeue. The
+	// predicate restricts the watch to the Gateway namespace, the only place a
+	// user-provided secret is read from. The OIDC client Secret in app namespaces
+	// is handled by Owns above; this watch is for the unowned, user-managed secret.
+	b = b.Watches(
+		&corev1.Secret{},
+		handler.EnqueueRequestsFromMapFunc(r.secretToNebariApp),
+		builder.WithPredicates(predicate.NewPredicateFuncs(func(obj client.Object) bool {
+			return obj.GetNamespace() == constants.GatewayNamespace
+		})),
+	)
+
+	return b.Complete(r)
+}
+
+// secretToNebariApp maps a user-provided TLS Secret in the Gateway namespace to
+// any NebariApp that references it via routing.tls.secretName. The NebariApp list
+// is served from the controller cache, so the per-event cost is negligible.
+func (r *NebariAppReconciler) secretToNebariApp(ctx context.Context, obj client.Object) []reconcile.Request {
+	var apps appsv1.NebariAppList
+	if err := r.List(ctx, &apps); err != nil {
+		return nil
+	}
+
+	var requests []reconcile.Request
+	for i := range apps.Items {
+		app := &apps.Items[i]
+		if app.Spec.Routing == nil || app.Spec.Routing.TLS == nil {
+			continue
+		}
+		if app.Spec.Routing.TLS.SecretName == obj.GetName() {
+			requests = append(requests, reconcile.Request{
+				NamespacedName: types.NamespacedName{Name: app.Name, Namespace: app.Namespace},
+			})
+		}
+	}
+	return requests
 }
 
 // certificateToNebariApp maps a cert-manager Certificate to the NebariApp that owns it
--- a/internal/controller/reconcilers/auth/providers/keycloak.go
+++ b/internal/controller/reconcilers/auth/providers/keycloak.go
@@ -37,14 +37,17 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 // KeycloakProvider implements the OIDCProvider interface for Keycloak.
 type KeycloakProvider struct {
 	Client client.Client
+	Scheme *runtime.Scheme
 	Config config.KeycloakConfig
 }
 
@@ -728,6 +731,17 @@ func (p *KeycloakProvider) storeClientSecret(ctx context.Context, nebariApp *app
 		Data: secretData,
 	}
 
+	// Set the NebariApp as the controller owner so the secret is garbage-collected
+	// with the NebariApp and so the controller's Owns(&corev1.Secret{}) watch maps
+	// secret events back to this NebariApp. The secret lives in the NebariApp's own
+	// namespace, so the owner reference is valid. Best-effort: a nil Scheme (e.g. in
+	// tests that don't wire one) skips the owner reference rather than failing.
+	if p.Scheme != nil {
+		if err := controllerutil.SetControllerReference(nebariApp, secret, p.Scheme); err != nil {
+			return fmt.Errorf("failed to set controller reference on OIDC client secret: %w", err)
+		}
+	}
+
 	// Check if secret exists
 	existingSecret := &corev1.Secret{}
 	err := p.Client.Get(ctx, types.NamespacedName{Name: secretName, Namespace: nebariApp.Namespace}, existingSecret)
@@ -739,8 +753,10 @@ func (p *KeycloakProvider) storeClientSecret(ctx context.Context, nebariApp *app
 		return fmt.Errorf("failed to check for existing secret: %w", err)
 	}
 
-	// Update existing secret
+	// Update existing secret data and ensure the owner reference is set (so secrets
+	// created before owner references were applied get adopted on the next sync).
 	existingSecret.Data = secret.Data
+	existingSecret.OwnerReferences = secret.OwnerReferences
 	return p.Client.Update(ctx, existingSecret)
 }
 
--- a/internal/controller/watches_test.go
+++ b/internal/controller/watches_test.go
@@ -1,0 +1,93 @@
+/*
+Copyright 2026, OpenTeams.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"context"
+	"reflect"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	appsv1 "github.com/nebari-dev/nebari-operator/api/v1"
+	"github.com/nebari-dev/nebari-operator/internal/controller/utils/constants"
+)
+
+// TestSecretToNebariApp covers the map function backing the user-provided TLS
+// secret watch (see #115): a secret enqueues only the NebariApp(s) that
+// reference it via routing.tls.secretName, and unrelated secrets enqueue
+// nothing. Namespace filtering is handled by the watch predicate, not the map
+// function, so it is not exercised here.
+func TestSecretToNebariApp(t *testing.T) {
+	scheme := runtime.NewScheme()
+	_ = appsv1.AddToScheme(scheme)
+	_ = corev1.AddToScheme(scheme)
+
+	appWithSecret := &appsv1.NebariApp{
+		ObjectMeta: metav1.ObjectMeta{Name: "app-a", Namespace: "ns-a"},
+		Spec: appsv1.NebariAppSpec{
+			Routing: &appsv1.RoutingConfig{TLS: &appsv1.RoutingTLSConfig{SecretName: "my-tls"}},
+		},
+	}
+	appNoSecretName := &appsv1.NebariApp{
+		ObjectMeta: metav1.ObjectMeta{Name: "app-b", Namespace: "ns-b"},
+		Spec: appsv1.NebariAppSpec{
+			Routing: &appsv1.RoutingConfig{TLS: &appsv1.RoutingTLSConfig{}},
+		},
+	}
+	appNoRouting := &appsv1.NebariApp{
+		ObjectMeta: metav1.ObjectMeta{Name: "app-c", Namespace: "ns-c"},
+	}
+
+	c := fake.NewClientBuilder().WithScheme(scheme).
+		WithObjects(appWithSecret, appNoSecretName, appNoRouting).Build()
+	r := &NebariAppReconciler{Client: c}
+
+	tests := []struct {
+		name       string
+		secretName string
+		want       []reconcile.Request
+	}{
+		{
+			name:       "secret referenced by a NebariApp enqueues that app",
+			secretName: "my-tls",
+			want:       []reconcile.Request{{NamespacedName: types.NamespacedName{Name: "app-a", Namespace: "ns-a"}}},
+		},
+		{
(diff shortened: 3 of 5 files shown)


internal/controller/reconcilers/auth/providers/keycloak.go
```diff
@@ -739,8 +753,10 @@ func (p *KeycloakProvider) storeClientSecret(ctx context.Context, nebariApp *app
 		return fmt.Errorf("failed to check for existing secret: %w", err)
 	}
 
-	// Update existing secret
+	// Update existing secret data and ensure the owner reference is set (so secrets
+	// created before owner references were applied get adopted on the next sync).
 	existingSecret.Data = secret.Data
+	existingSecret.OwnerReferences = secret.OwnerReferences
```
dcmcand: 🛑 **Blocker:** this replaces the whole `OwnerReferences` slice, and the `AlreadyOwnedError` guard that should have prevented a hostile takeover is unreachable, so any owner reference already on the Secret is silently discarded.

Call `controllerutil.SetControllerReference(nebariApp, existingSecret, ...)` on the fetched object here instead - that lets the real guard fire - and preserve any additional non-controlling owner references rather than overwriting the slice.

At line 740 `SetControllerReference` is applied to `secret`, the fresh literal built at `:720-732`, which never carries owner references from the cluster. Its protective check is `if existing := metav1.GetControllerOf(controlled); existing != nil && !referSameObject(...)` (controller-runtime v0.22.4, `pkg/controller/controllerutil/controllerutil.go:98`), so on a freshly built object it can never fire; `existingSecret`, fetched at `:747`, is never passed in.

Reproduced against a fake client - a pre-existing Secret at the derived name carrying a controller owner reference to an `ExternalSecret` plus one additional non-controller owner:

```
storeClientSecret err = <nil>
resulting ownerReferences count = 1
  -> reconcilers.nebari.dev/v1/NebariApp name=test-app uid=nebariapp-uid controller=true
VERIFIED: owner reference uid="foreign-controller-uid" was silently dropped
VERIFIED: owner reference uid="extra-owner-uid" was silently dropped
```

Both gone, `err == nil`, no event, no log. Kubernetes GC only tracks what is in `ownerReferences` in etcd, so the previous owner's cascade silently stops working, while deleting this NebariApp now cascade-deletes a credential-bearing Secret whose lifecycle nothing verified belonged to it. The comment above describes the intended nil-to-one-reference adoption; the code does something broader.

This line is also untested: removing it outright leaves the entire `internal/controller/reconcilers/auth/...` suite green.

internal/controller/reconcilers/auth/providers/keycloak.go
```diff
@@ -728,6 +731,17 @@ func (p *KeycloakProvider) storeClientSecret(ctx context.Context, nebariApp *app
...
+	// Set the NebariApp as the controller owner so the secret is garbage-collected
+	// with the NebariApp and so the controller's Owns(&corev1.Secret{}) watch maps
+	// secret events back to this NebariApp. The secret lives in the NebariApp's own
+	// namespace, so the owner reference is valid. Best-effort: a nil Scheme (e.g. in
+	// tests that don't wire one) skips the owner reference rather than failing.
+	if p.Scheme != nil {
+		if err := controllerutil.SetControllerReference(nebariApp, secret, p.Scheme); err != nil {
+			return fmt.Errorf("failed to set controller reference on OIDC client secret: %w", err)
+		}
+	}
```
dcmcand: 💡 **Suggestion:** drop the `Scheme` field and this nil guard, and read the scheme off the client instead - `p.Client.Scheme()`.

`client.Client` exposes `Scheme()` (v0.22.4, `pkg/client/interfaces.go:174`) and this repo already uses that idiom at `internal/controller/reconcilers/auth/rbac.go:60,85`. `cmd/operator/main.go:215` reverts with it.

The guard is unreachable in production - `mgr.GetScheme()` is never nil - and it is not the test convenience the comment claims: the existing table-driven `TestKeycloakProvider_StoreClientSecret` omits `Scheme` (`keycloak_test.go:457-460`), so every one of its cases including the update case currently runs the no-owner-reference path and executes `existingSecret.OwnerReferences = nil`. Removing the field makes those tests cover owner references for free, and removes a landmine where a future caller silently regresses to the exact bug #30 closed.

internal/controller/nebariapp_controller.go
```diff
@@ -382,22 +386,70 @@ func (r *NebariAppReconciler) cleanup(ctx context.Context, nebariApp *appsv1.Neb
...
+		Named("nebariapp").
+		// Watch the resources the operator owns so that external edits or
+		// deletions trigger prompt reconciliation (re-creation / drift repair)
+		// instead of waiting up to the periodic requeue interval. Owned objects
+		// are mapped back to their NebariApp via controller owner references:
+		// HTTPRoutes and SecurityPolicies set them at construction; the OIDC
+		// client Secret now sets one in storeClientSecret.
+		Owns(&gatewayv1.HTTPRoute{}).
+		Owns(&egv1alpha1.SecurityPolicy{}).
+		Owns(&corev1.Secret{})
```
dcmcand: 💡 **Suggestion:** extend `Owns` to every type the operator controller-owns - Role and RoleBinding are missing.

Add `Owns(&rbacv1.Role{})` and `Owns(&rbacv1.RoleBinding{})` (needs an `rbacv1 "k8s.io/api/rbac/v1"` import). RBAC already grants `watch` on both at `nebariapp_controller.go:71-72`, so no marker change is needed.

Full inventory of controller owner references in non-test code: HTTPRoute (`routing/httproute.go:217`, `:458`), SecurityPolicy (`auth/reconciler.go:390`), Secret (`keycloak.go:740`, new here) are covered; Role (`auth/rbac.go:60`) and RoleBinding (`auth/rbac.go:85`) are not. Deleting the RoleBinding revokes the app ServiceAccount's read access to its own OIDC secret - comparable blast radius to the HTTPRoute you did watch - and is repaired only on the 1m requeue. The Certificate stays out correctly, being cross-namespace and label-tracked by design (`tls/reconciler.go:273`).

internal/controller/nebariapp_controller.go
```diff
@@ -382,22 +386,70 @@ func (r *NebariAppReconciler) cleanup(ctx context.Context, nebariApp *appsv1.Neb
...
 	}
 
-	return builder.Complete(r)
+	// Watch user-provided TLS secrets in the Gateway namespace. When a NebariApp
+	// references routing.tls.secretName for a secret that does not yet exist (or
+	// has the wrong type), TLSReady stays False; creating or fixing the secret
+	// should flip it promptly rather than on the next periodic requeue. The
+	// predicate restricts the watch to the Gateway namespace, the only place a
+	// user-provided secret is read from. The OIDC client Secret in app namespaces
+	// is handled by Owns above; this watch is for the unowned, user-managed secret.
```
dcmcand: ❤️ **Kudos:** this comment explicitly separates the owned OIDC Secret in app namespaces from the unowned user-managed secret in the Gateway namespace.

Two watches on the same GVK with different semantics is exactly what a future reader collapses into one. This note is what prevents that, and it measurably sped up tracing the real event routing against the controller-runtime source.

internal/controller/nebariapp_controller.go
```diff
@@ -382,22 +386,70 @@ func (r *NebariAppReconciler) cleanup(ctx context.Context, nebariApp *appsv1.Neb
...
+	b = b.Watches(
+		&corev1.Secret{},
+		handler.EnqueueRequestsFromMapFunc(r.secretToNebariApp),
+		builder.WithPredicates(predicate.NewPredicateFuncs(func(obj client.Object) bool {
+			return obj.GetNamespace() == constants.GatewayNamespace
+		})),
```
dcmcand: 💡 **Suggestion:** this predicate is decision logic living inside a side-effecting wrapper, which makes it structurally unreachable from tests. Extract it to a named pure function so `SetupWithManager` stays pure wiring:

```go
func inGatewayNamespace(obj client.Object) bool { return obj.GetNamespace() == constants.GatewayNamespace }
```

then `builder.WithPredicates(predicate.NewPredicateFuncs(inGatewayNamespace))`.

`secretToNebariApp` below gets this split right - inputs to `[]reconcile.Request`, no side effects - and that is exactly why it is testable and tested. This closure is the same kind of logic, but `TestSecretToNebariApp` calls the map function directly and never touches the builder, so deleting the predicate entirely leaves both subtests green. #115's acceptance criteria asks specifically for the namespace filter.

The coverage profile shows the split cleanly: on this head, `go test -coverprofile` reports this closure with execution count **0** while every branch of `secretToNebariApp` is count 1. Worth noting for the other half of the gap, though - coverage will *not* flag `keycloak.go:759`, which reports count **1** because the existing update-path test executes it and simply never asserts on it. Coverage finds unreachable logic; mutation is what finds unasserted logic.

internal/controller/nebariapp_controller.go
```diff
@@ -382,22 +386,70 @@ func (r *NebariAppReconciler) cleanup(ctx context.Context, nebariApp *appsv1.Neb
...
+}
+
+// secretToNebariApp maps a user-provided TLS Secret in the Gateway namespace to
+// any NebariApp that references it via routing.tls.secretName. The NebariApp list
+// is served from the controller cache, so the per-event cost is negligible.
+func (r *NebariAppReconciler) secretToNebariApp(ctx context.Context, obj client.Object) []reconcile.Request {
+	var apps appsv1.NebariAppList
+	if err := r.List(ctx, &apps); err != nil {
+		return nil
+	}
```
dcmcand: 💡 **Suggestion:** this `List` error is swallowed with no log, so a failed lookup drops the event with zero operator-visible signal.

Log it before returning:

```go
	if err := r.List(ctx, &apps); err != nil {
		logf.FromContext(ctx).Error(err, "failed to list NebariApps for secret watch", "secret", obj.GetName())
		return nil
	}
```

(`logf` is already imported.) This is the only watch in the file whose event delivery depends on a call succeeding, so a transient API error or RBAC drift presents as "TLSReady never flips promptly" with nothing anywhere to explain it - the app still recovers on the periodic requeue, but the fast path failing is invisible. Every other error path in this file logs (`:94`, `:112`, `:141`, `:160`, `:190`, and ten more); `certificateToNebariApp` just below is the one legitimate exception, since it discards its context and has no error to report. The coverage profile confirms this branch is never executed by any test.

internal/controller/watches_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+	appsv1 "github.com/nebari-dev/nebari-operator/api/v1"
+	"github.com/nebari-dev/nebari-operator/internal/controller/utils/constants"
+)
+
+// TestSecretToNebariApp covers the map function backing the user-provided TLS
+// secret watch (see #115): a secret enqueues only the NebariApp(s) that
+// reference it via routing.tls.secretName, and unrelated secrets enqueue
+// nothing. Namespace filtering is handled by the watch predicate, not the map
+// function, so it is not exercised here.
```
dcmcand: ❤️ **Kudos:** stating outright that namespace filtering "is not exercised here", rather than letting the table-driven shape imply coverage it does not have.

Nothing required that disclosure, and it is what let the gap be found by inspection instead of by surprise - the coverage profile independently confirms the predicate closure at `nebariapp_controller.go:423-425` is never executed.

dcmcand (changes requested):
**Verdict:** Not approvable as-is - one correctness Blocker in `storeClientSecret` plus three documentation edits, one of which is a live falsehood in user-facing troubleshooting.

The Blocker and the `Scheme`-field suggestion inline collapse into a single small change. Otherwise this is solid, idiomatic controller-runtime work with unusually good comments, and it quietly fixes a Secret leak the docs already claimed was handled.

Verified at `5870e3c`: `go build`, `go vet`, `gofmt -l` clean; `go test ./internal/controller/reconcilers/auth/...` and `TestSecretToNebariApp` pass; `make manifests` produces zero diff on `config/rbac/role.yaml`. The `internal/controller` envtest suite fails identically at this head and at the base `f7a2e9d` (`no kind is registered for the type v1alpha1.SecurityPolicy`), so the "also fails on clean `main`" claim in the description checks out. The owner-reference finding was reproduced against a fake client and mutation-tested. Nothing was run against a live cluster, so watch delivery is verified from the controller-runtime v0.22.4 source rather than empirically.

## Documentation

`AGENTS.md:270` requires contract changes to be recorded in the same PR, and `:277` designates `docs/reconcilers/` "the source of truth for reconciler behavior".

🛑 **Blocker:** `docs/troubleshooting.md:35` states "Re-readiness today depends on the periodic requeue (~30 seconds)", which this PR makes false.
Replace that sentence with the watch-driven behaviour.
A user hitting `UserProvidedSecretNotFound` is told to expect ~30s when it is now sub-second. This is the exact sentence #115 exists to invalidate.

🛑 **Blocker:** `docs/reconcilers/authentication.md` does not record that the OIDC client Secret's deletion semantics changed.
Add an "Owner reference: NebariApp (for garbage collection)" line to the Secret block at `:123`, matching the one `:161` already gives SecurityPolicy, and list the Secret alongside SecurityPolicy in step 4 of the Cleanup Process at `:298`.
`CleanupAuth` (`internal/controller/reconcilers/auth/reconciler.go:266-300`) deletes the Keycloak client but never the Secret, so before this PR the Secret survived NebariApp deletion; now it cascades. Nothing in the doc is false - it is silent - so the basis here is the project's own same-PR rule plus this being operator-visible deletion behaviour for a credential-bearing object.

🛑 **Blocker:** `AGENTS.md:204` becomes false on merge - "`SetupWithManager` watches the primary `NebariApp` plus cert-manager `Certificate` objects, mapped back ... via the `nebari.dev/nebariapp-name` / `nebari.dev/nebariapp-namespace` labels". There are now three mapping mechanisms: controller owner reference, label, and a `spec.routing.tls.secretName` scan.
Rewrite the sentence to enumerate the four watch sources and the three mappings.
You could not have seen this - `AGENTS.md` landed on `main` in 58a1074 after this branch point, and the branch is 12 commits behind (it still merges cleanly). CI will not catch it either: `hack/gen-agents.sh` regenerates only the machine-owned make-targets block, so the prose is unchecked.

💡 **Suggestion:** `AGENTS.md:275` points contributors at five files in `docs/design/` - including `epic-routing-securitypolicy.md`, the design doc for this very epic - and that directory does not exist in the repo (`git ls-files 'docs/design/*'` is empty).
Worth a follow-up issue: either commit the docs or drop the pointer.
Predates this PR, but it means the instruction to "read the matching one before touching auth, routing..." currently sends a contributor to nothing.

💡 **Suggestion:** `docs/reconcilers/validation.md:197` claims namespace-label and Service creation "trigger reconciliation via watches". There is no Service or Namespace watch even after this PR, so the line is false before and after.
Follow-up issue, or fix in passing while the watch list is already being edited.

## Elsewhere in the tree

💡 **Suggestion:** `internal/controller/reconcilers/routing/httproute.go:104` and `:343` - both HTTPRoute update paths assign only `existingRoute.Spec = desiredRoute.Spec` and never restore `OwnerReferences`, so a route whose owner reference is stripped is never re-adopted.
Also assign `existingRoute.OwnerReferences = desiredRoute.OwnerReferences` in both, or convert them to the `CreateOrUpdate`-with-mutate shape the other reconcilers use.
This predates the PR, but the PR makes it load-bearing: `Owns` resolves by controller owner reference, so a stripped reference permanently silences the new watch for that route. It degrades to the 1m requeue rather than breaking, which is why this is not a Blocker. You already solved exactly this case for the Secret.

💡 **Suggestion:** the description's cache note ("`Owns(&corev1.Secret{})` makes the manager cache Secrets in watched namespaces") scopes the follow-up wrong, and it will outlive the PR.
Reword to say the informer already exists and `Owns` only makes it eager.
`ctrl.Options` at `cmd/operator/main.go:170-188` sets no `Cache` or `Client.Cache` fields, so `DisableFor` is empty and every structured Secret read already routes through the cache today - `tls/reconciler.go:560`, `auth/reconciler.go:324`, `keycloak.go:166` all do this on `main`. If you do want to bound it later, the lever is `cache.Options.ByObject`, not a watch predicate; predicates run in the handler, after the object is already cached.

❓ **Question:** where does #30's "integration test confirms deleting an owned HTTPRoute triggers re-creation" actually land?
Deferring it is reasonable, but the only home for it is broken and invisible: the `internal/controller` envtest suite fails on the base branch, and `Makefile:68` greps `internal/controller$` out of `make test`, so CI never runs it. I found no open issue tracking either problem. Is that tracked somewhere I could not see, or should this PR file it?

---
**Key:** 🛑 Blocker (must fix before merge) · 💡 Suggestion (improvement, not blocking) · ❓ Question (needs author input) · ❤️ Kudos

PR #111, opened 2026-04-27
fix(auth/keycloak): use external Keycloak URL for browser-facing OIDC endpoints

dcmcand:
## Summary
- Split SecurityPolicy OIDC endpoint overrides by audience: Token stays on the in-cluster URL (Envoy back-channel), Authorization and EndSession move to the public URL (browser front-channel).
- Adds an `externalRealmURL()` helper mirroring `internalRealmURL()`.
- When `KEYCLOAK_EXTERNAL_URL` is unset, leaves Authorization/EndSession unset; Envoy Gateway then triggers OIDC discovery against the in-cluster issuer. Discovery only produces public URLs if Keycloak's own `frontendUrl` is configured. Operator now logs a startup warning so this trap is visible.
- Updated `TestKeycloakProvider_GetEndpointOverrides` with five cases covering the split, trailing-slash normalization, and the no-ExternalURL fallback.

## Why
On a clean NIC deploy, browsers hitting any NebariApp protected by Keycloak OAuth2 are redirected to `http://keycloak-keycloakx-http.keycloak.svc.cluster.local:8080` - which the browser cannot resolve. The OAuth2 flow dead-ends and no UI is reachable.

Surfaced during fresh-install validation of the Nebari LLM Serving Pack (nebari-dev/nebari-llm-serving-pack#65), which loads the key-manager UI as a NebariApp.

Closes #110

## Follow-ups (deliberately out of scope here)
- #112 - Should `SecurityPolicy.spec.oidc.provider.Issuer` also track `KEYCLOAK_EXTERNAL_URL` to match the `iss` claim in tokens issued by Keycloak with `frontendUrl` configured?
- #113 - Add a reconciler-level test asserting Token on in-cluster URL and Authorization+EndSession on public URL end-to-end through `buildSecurityPolicySpec`.

## Test plan
- [x] `go test ./internal/controller/reconcilers/auth/providers/...` passes - five cases covering the split, trailing-slash normalization, no-ExternalURL fallback
- [x] `go test ./...` no regressions vs main (pre-existing `TestControllers` HTTPRoute scheme failure unrelated)
- [x] `go build ./...` clean
- [ ] Validation against the llmd-validate cluster: build the operator image, swap it in via a temporary override on NIC's nebari-operator Application, confirm the rendered SecurityPolicy now uses `https://keycloak.<base>` for Authorization+EndSession, confirm browser login completes

PR #144, opened 2026-06-12
docs(auth): SecurityPolicy issuer stays in-cluster — investigation result for #112, with tests

### LINKED ISSUE #112: OIDC SecurityPolicy.Issuer field stays on in-cluster URL; investigate iss-claim mismatch with Keycloak frontendUrl deployments

### LINKED ISSUE #113: Add reconciler test asserting SecurityPolicy.spec.oidc.provider has Token on in-cluster URL and Authorization on public URL

tylerpotts:
## Summary

Answers the investigation in https://github.com/nebari-dev/nebari-operator/issues/112: **`SecurityPolicy.spec.oidc.provider.issuer` should NOT track `KEYCLOAK_EXTERNAL_URL`.** The in-cluster issuer is correct and the iss-claim mismatch with Keycloak `frontendUrl` deployments is harmless by construction. This PR records that decision in code comments and docs, and adds tests locking in the invariant that makes it safe.

## Investigation findings (verified against Envoy Gateway v1.6.3 and Envoy source)

1. **Envoy Gateway uses the issuer only at control-plane time**, to fetch `/.well-known/openid-configuration` — and only when `authorizationEndpoint` or `tokenEndpoint` is not explicitly set (`internal/gatewayapi/securitypolicy.go`, `buildOIDCProvider`). When `KEYCLOAK_EXTERNAL_URL` is configured the operator sets both, so EG never contacts the issuer and the IR passed to the data plane does not carry it.
2. **The Envoy `oauth2` filter never validates `iss`.** Its config proto (`api/envoy/extensions/filters/http/oauth2/v3/oauth.proto`) has no issuer field at all; the filter's only JWT decode is an unverified `exp` read for cookie lifetimes (`source/extensions/filters/http/oauth2/filter.cc`). This holds through Envoy 1.39-dev.
3. **EG does not enforce the OIDC Discovery issuer-match rule** (OpenID Connect Discovery 1.0 §4.3): it unmarshals only the endpoint fields from the discovery document and ignores `issuer`. So even the discovery fallback path works when Keycloak returns a public issuer on the internal endpoint.
4. **Consumers that DO strictly enforce `iss`** — `spec.jwt` providers (Envoy `jwt_authn`), e.g. the one the LLM serving pack's operator generates — already receive the **public** issuer via the client Secret's `issuer-url` key, populated from `GetExternalIssuerURL()`. That key is empty when `KEYCLOAK_EXTERNAL_URL` is unset, so strict-iss consumers require it to be set.
5. **Flipping the issuer to the external URL would be a pure regression**: no data-plane effect, but the discovery fallback would then be fetched from the envoy-gateway controller pod against the public URL — reintroducing the public TLS-chain trust and hairpin-routing problems that https://github.com/nebari-dev/nebari-operator/pull/111 deliberately avoided for the token endpoint.

## Changes

- `GetIssuerURL` / `GetEndpointOverrides` doc comments record the verified semantics and the both-endpoints-set invariant.
- `TestKeycloakProvider_GetIssuerURL`: new case asserting the issuer stays in-cluster when `ExternalURL` is set.
- `TestBuildSecurityPolicySpec_KeycloakIssuerAndEndpointSplit`: reconciler-level test through `buildSecurityPolicySpec` with the real `KeycloakProvider` — issuer and token endpoint on the in-cluster host, authorization/end-session on the external URL, plus the no-ExternalURL fallback. This is the reconciler-level endpoint-split test requested in https://github.com/nebari-dev/nebari-operator/issues/113.
- `docs/reconcilers/authentication.md`: new **Issuer Semantics (Keycloak)** section.

Closes https://github.com/nebari-dev/nebari-operator/issues/112
Closes https://github.com/nebari-dev/nebari-operator/issues/113

## Test plan

- [x] `go test ./internal/controller/reconcilers/auth/...` passes; new tests verified to execute (`-v -run`)
- [x] `golangci-lint run` — 0 issues
- [x] `go test ./...` — no regressions vs main (pre-existing `TestControllers` HTTPRoute scheme failure reproduced on main, unrelated)

### DIFF
--- a/docs/reconcilers/authentication.md
+++ b/docs/reconcilers/authentication.md
@@ -193,6 +193,33 @@ spec:
 - `redirectURL`: Defaults to `https://{hostname}/oauth2/callback`
 - `scopes`: Uses `spec.auth.scopes` or defaults to `["openid", "profile", "email"]`
 
+### Issuer Semantics (Keycloak)
+
+For Keycloak, `provider.issuer` is always the **in-cluster** realm URL, even when
+`KEYCLOAK_EXTERNAL_URL` is set and Keycloak (with a public `frontendUrl`) stamps the
+**public** URL into every token's `iss` claim. This mismatch is deliberate and harmless
+([investigation in #112](https://github.com/nebari-dev/nebari-operator/issues/112)):
+
+- Envoy Gateway uses the issuer **only at control-plane time** to fetch
+  `/.well-known/openid-configuration`, and only when `authorizationEndpoint` or
+  `tokenEndpoint` is not explicitly set. When `KEYCLOAK_EXTERNAL_URL` is configured the
+  operator sets both, so the issuer URL is never contacted and never reaches the data
+  plane.
+- The Envoy `oauth2` filter has no issuer field in its configuration and never inspects
+  the token's `iss` claim (its only JWT decode is an unverified `exp` read for cookie
+  lifetimes).
+- Envoy Gateway does not enforce the OIDC Discovery issuer-match rule: it reads only the
+  endpoint fields from the discovery document and ignores its `issuer` field.
+- Consumers that **do** strictly validate `iss` — e.g. a SecurityPolicy `spec.jwt`
+  provider configured by a downstream pack — must use the public issuer. They read it
+  from the client Secret's `issuer-url` key, which is populated from
+  `GetExternalIssuerURL()` (the public URL). That key is empty when
+  `KEYCLOAK_EXTERNAL_URL` is unset, so strict-`iss` consumers require it to be set.
+
+Keeping the issuer in-cluster means the discovery fallback (used when
+`KEYCLOAK_EXTERNAL_URL` is unset) runs from the envoy-gateway controller pod over
+cluster DNS, avoiding public TLS-chain trust and hairpin-routing requirements.
+
 **On Failure:**
 - Event: `Warning` with reason `SecurityPolicyFailed`
 - Condition: `AuthReady=False` with reason `SecurityPolicyFailed`
--- a/internal/controller/reconcilers/auth/providers/keycloak.go
+++ b/internal/controller/reconcilers/auth/providers/keycloak.go
@@ -71,7 +71,29 @@ func (p *KeycloakProvider) externalRealmURL() string {
 }
 
 // GetIssuerURL returns the internal cluster URL for the Keycloak realm.
-// Envoy uses this to fetch OIDC configuration from within the cluster.
+//
+// This deliberately stays on the in-cluster URL even when Keycloak issues
+// tokens with a public `iss` claim (frontendUrl / KC_HOSTNAME configured).
+// Investigated in https://github.com/nebari-dev/nebari-operator/issues/112:
+//
+//   - Envoy Gateway uses SecurityPolicy.spec.oidc.provider.issuer only at
+//     control-plane time, to fetch /.well-known/openid-configuration — and
+//     only when authorizationEndpoint or tokenEndpoint is not explicitly set
+//     (gateway v1.6.3 internal/gatewayapi/securitypolicy.go). The issuer is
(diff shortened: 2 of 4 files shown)


viniciusdc (approved):
Reviewed the contents, not just the summary - this is sound.

- Comment + docs + test only; no behavioral change. The issuer/endpoint-split behavior shipped in #111; this documents the decision and locks it with tests, so there's zero regression risk.
- `TestBuildSecurityPolicySpec_KeycloakIssuerAndEndpointSplit` drives the real `KeycloakProvider` through `buildSecurityPolicySpec` and covers both the external-set path and the discovery-fallback path. Expected values are literals, not re-derived from the provider, so a URL-construction regression is actually caught. `GetIssuerURL` gets the matching in-cluster-issuer case.
- Docs and code comments match the source.

I also re-verified the underlying claims against the EG upgrade target (v1.8.2) and latest Envoy: `buildOIDCProvider` and the OIDC IR are unchanged, and the `oauth2` filter proto still has no issuer field - so this does not reopen when the EG bump lands (details on #112).

One non-blocking nit: the `GetIssuerURL` comment hard-pins "gateway v1.6.3"; it'll read stale after the EG upgrade even though the behavior is identical on v1.8.2. Fine to leave or tweak in passing.

PR #150, opened 2026-07-02
Add nebari-app library Helm chart for NebariApp CR

pmeier:
## Reference Issues or PRs

Closes #148.

## What does this implement/fix?

_Put a `x` in the boxes that apply_

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds a feature)
- [ ] Breaking change (fix or feature that would cause existing features not to work as expected)
- [ ] Documentation Update
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build related changes
- [ ] Other (please describe):

## Testing

- [x] Did you test the pull request locally?
- [x] Did you add new tests?

## Documentation

`charts/nebari-app/README.md`


charts/nebari-app/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Install as a dependency
+
+Add the chart to your consumer chart's `Chart.yaml`:
+
+```yaml
+dependencies:
+  - name: nebari-app
+    version: 0.1.0
+    repository: https://nebari.dev/charts
```
pmeier:
```suggestion
    repository: oci://quay.io/nebari/charts
```

dcmcand: **Blocker.** This URL does not serve a Helm repository:

```
$ curl -o /dev/null -w '%{http_code}' https://nebari.dev/charts            -> 404
$ curl -o /dev/null -w '%{http_code}' https://nebari.dev/charts/index.yaml -> 404
$ helm repo add nebari https://nebari.dev/charts
Error: looks like "https://nebari.dev/charts" is not a valid chart repository or cannot be
reached: failed to fetch https://nebari.dev/charts/index.yaml : 404 Not Found
```

`helm dependency build` on the snippet fails the same way. The host resolves (Netlify) but has no `/charts` path, `nebari-docs`'s `netlify.toml` has no redirect for it, and a GitHub-wide code search for `nebari.dev/charts` returns zero hits.

This matches your own suggestion on this PR, which is not yet applied. `oci://quay.io/nebari/charts` is what `chat-pack`'s `Chart.lock` uses for `ravnar` and what `docs/maintainers/release-process.md:153` documents. The gh-pages raw index from `helm-repository`'s README works too if you would rather point at the classic repo.

```suggestion
    repository: oci://quay.io/nebari/charts
```

charts/nebari-app/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+The template enforces these required fields via Helm's `required` function (render aborts if any is missing or empty):
+
+- `metadata.name`
+- `metadata.namespace`
+- `spec.hostname`
+- `spec.service.name`
+- `spec.service.port`
+
+All other validation (the rest of the `NebariAppSpec` schema) happens API-server-side at apply time. Pipe `helm template` output through `kubectl --dry-run=client -f -` in CI to catch schema errors before deployment.
```
dcmcand: **Blocker.** Two problems here. The command is not runnable as written, which is the minor half:

```
$ cat cr.yaml | kubectl --dry-run=client -f -
Error: unknown flag: --dry-run
```

The substantive half is that even corrected to `kubectl apply --dry-run=client`, it cannot validate a NebariApp. Tested against a real apiserver (envtest, k8s 1.34.1) with this PR's CRD installed, on three invalid resources: unknown field `spec.bogusField`, `spec.service.port: "not-an-integer"`, and missing required `spec.hostname`.

| kubectl | `apply --dry-run=client` (default `--validate=strict`) | `apply --dry-run=server` |
|---|---|---|
| 1.30 / 1.31 / 1.32 / 1.33 / 1.34 / 1.35 / 1.36 | all 3 report `created (dry run)`, **exit 0** | all 3 caught, exit 1 |

`--validate=strict` and `--validate=true` make no difference, and duplicate YAML keys pass too. Without a cluster it does not run at all (`failed to download openapi`), so it is not an offline check either. `kubectl apply --help` explains why: validation is server side, and `--dry-run=client` sends no request.

A check that reports success on invalid input is worse than no check, since CI will look green.

```suggestion
All other validation (the rest of the `NebariAppSpec` schema) happens API-server-side at apply time. Pipe `helm template` output through `kubectl apply --dry-run=server -f -` in CI to catch schema errors before deployment. This requires a cluster with the NebariApp CRD installed; `--dry-run=client` cannot validate custom resources and will report success on invalid input.
```

Makefile
```diff
@@ -199,7 +200,31 @@ helm-chart-version: ## Update Helm chart version and appVersion (requires VERSIO
...
+	@for c in static computed multi; do \
+		helm template t test/fixture --set cases.$$c.enabled=true >/dev/null \
+			|| { echo >&2 "case $$c failed to render"; exit 1; }; \
+		echo "  case $$c: ok"; \
+	done
+	@for f in spec.hostname metadata.name metadata.namespace spec.service.name spec.service.port; do \
+		helm template t test/fixture --set cases.static.enabled=true --set cases.static.$$f= >/dev/null 2>&1 \
+			&& { echo >&2 "expected required failure for $$f"; exit 1; }; \
+		echo "  required guard $$f: ok"; \
+	done
```
dcmcand, line 253: **Blocker.** The chart is correct today. I rendered all three cases and parsed the output: two valid `NebariApp` docs from the multi case, labels propagated, merge precedence right. This is about the harness.

Replacing the library template with nothing but its five `required` calls, so it emits no resource at all:

```
$ make helm-test
  required guard spec.service.port: ok
nebari-app chart tests passed

$ helm template t test/fixture --set cases.static.enabled=true | wc -c
1
```

One byte of output, suite green. Seven more mutations also stay green: `apiVersion` to `totally.bogus.example.com/v99`, `kind` to `NotAnAppAtAll`, deleting the whole `metadata:` block, deleting the whole `spec:` block, `nindent 2` to `nindent 0` (which makes `spec` null and promotes every spec field to a top-level key), dropping caller-supplied labels, and swapping the `mergeOverwrite` argument order in the fixture.

The guard loop also does not localize failures to the library chart. Moving all five `required` calls out of the chart and into `test/fixture/templates/static.yaml` still prints all five `required guard X: ok` lines, because `>/dev/null 2>&1` discards the message and the loop only checks that *something* failed.

So the suite pins two things: that five `required` calls exist somewhere in the render path, and that the output parses as YAML. Nothing about the emitted document's identity, structure, or content, and the `mergeOverwrite` precedence rule at README:75 has no test at all.

Two fixes cover it. Golden-file compare the cases, which catches all eight mutations above (tab-indent as usual in the recipe):

```make
@for c in static computed multi; do \
    helm template t test/fixture --set cases.$$c.enabled=true \
        | diff -u test/fixture/testdata/$$c.golden - \
        || { echo >&2 "case $$c output drifted"; exit 1; }; \
    echo "  case $$c: ok"; \
done
```

And assert the guard's reason so it cannot pass for an unrelated render failure:

```make
helm template t test/fixture --set cases.static.enabled=true --set cases.static.$$f= 2>&1 >/dev/null \
    | grep -q "$$f is required" || { echo >&2 "expected required failure for $$f"; exit 1; }
```

Makefile
```diff
@@ -199,7 +200,31 @@ helm-chart-version: ## Update Helm chart version and appVersion (requires VERSIO
 	sed -i.bak "s/^version:.*/version: $(VERSION)/" dist/chart/Chart.yaml
 	sed -i.bak "s/^appVersion:.*/appVersion: \"$(APP_VERSION)\"/" dist/chart/Chart.yaml
 	rm -f dist/chart/Chart.yaml.bak
-	@echo "✅ Updated chart version to $(VERSION) and appVersion to $(APP_VERSION)"
+	sed -i.bak "s/^version:.*/version: $(VERSION)/" charts/nebari-app/Chart.yaml
+	sed -i.bak "s/^appVersion:.*/appVersion: \"$(APP_VERSION)\"/" charts/nebari-app/Chart.yaml
+	rm -f charts/nebari-app/Chart.yaml.bak
```
dcmcand, line 205: **Question.** Is it intended that `nebari-app`'s published version tracks the operator's release version?

`sync-chart/action.yml` resolves the version to publish by grepping `^version:` out of `charts/nebari-app/Chart.yaml`, and this rewrites that file during release. So the library chart gets a new published version on every operator release even when `_nebari-app.tpl` is untouched, and `appVersion` becomes the operator tag, which has no meaning for a library chart with no app.

Two knock-on effects, flagged separately on `README.md:14` and `test/fixture/Chart.yaml:9`. If independent versioning is what you want, dropping these three lines resolves both.

charts/nebari-app/README.md
```diff
@@ -0,0 +1,107 @@
...
+The chart exposes one named template, `nebari-app.nebariApp`, that acts as a pure function: callers pass `metadata` and `spec` dicts and the template renders a `NebariApp` resource. It is not installable on its own — it is consumed as a [chart dependency](https://helm.sh/docs/helm/helm_dependency/).
+
+## Install as a dependency
+
+Add the chart to your consumer chart's `Chart.yaml`:
+
+```yaml
+dependencies:
+  - name: nebari-app
+    version: 0.1.0
```
dcmcand: Given `Makefile:203-205`, this pin will not resolve from the Helm repository during the current `v0.1.0-alpha.N` series, since the published version tracks the tag. Verified empirically against a local repo serving only `nebari-app-0.1.0-alpha.21.tgz`:

```
Error: can't get a valid version for 1 subchart(s): "nebari-app" (..., version "0.1.0").
```

It becomes correct once a `v0.1.0` tag ships, and that path is real: `provenance-collector` is published at `['0.1.0', '0.1.0-alpha.2', '0.1.0-alpha.3', '0.1.0-alpha.4']`. So this is stale-for-now rather than permanently wrong. The `file://` snippet below at line 23 is unaffected, since it resolves against the in-tree `Chart.yaml`.

test/fixture/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v2
+name: fixture
+description: Test fixture consuming the nebari-app library chart.
+type: application
+version: 0.1.0
+appVersion: "0.1.0"
+dependencies:
+  - name: nebari-app
+    version: 0.1.0
```
dcmcand: An exact pin here breaks the fixture after any version bump:

```
$ sed -i 's/^version:.*/version: 0.1.0-alpha.21/' charts/nebari-app/Chart.yaml
$ make helm-test
Error: can't get a valid version for 1 subchart(s): "nebari-app"
  (repository "file://../../charts/nebari-app", version "0.1.0")
```

CI does not hit this, since it always runs on a clean checkout, but anyone running `make helm-chart-version` then `make helm-test` locally will.

```suggestion
    version: ">=0.1.0-0"
```

.github/workflows/test-chart.yml
```diff
@@ -55,6 +55,9 @@ jobs:
         run: |
           helm lint ./dist/chart
 
+      - name: Test nebari-app library chart
+        run: make helm-test
```
dcmcand: This runs in the kind-cluster job, but `make helm-test` needs no cluster. `build-pr.yml:214`'s "Helm Chart Tests" job is the more natural home, and it is where a maintainer would look for a chart test.

Makefile
```diff
@@ -186,7 +186,8 @@ helm-package: ## Package the Helm chart (run helm-chart first).
 	@command -v helm >/dev/null 2>&1 || { echo >&2 "helm is required but not installed. See https://helm.sh/docs/intro/install/"; exit 1; }
 	@if [ ! -d "dist/chart" ]; then echo "Error: dist/chart/ not found. Run 'make helm-chart' first."; exit 1; fi
 	helm package dist/chart --destination dist/
-	@echo "✅ Helm chart packaged in dist/"
+	helm package charts/nebari-app --destination dist/
```
dcmcand, line 189: This line is reachable from no pull-request path. `helm-package` appears exactly once in `.github/workflows/`, at `release.yml:153`, in a job triggered only by `on: release: types: [published]`, and no Makefile target lists it as a prerequisite. I confirmed it works locally (`helm package` handles `type: library` fine), but a regression here would surface mid-release rather than on a PR.

charts/nebari-app/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: nebari-app
+description: Library chart providing a NebariApp custom resource template.
+type: library
+version: 0.1.0
+appVersion: "0.1.0"
```
dcmcand, line 6: No `icon`, `home`, or `sources`; `helm lint` emits `[INFO] Chart.yaml: icon is recommended`. Worth knowing that `nebari-app` would be the first `type: library` chart and the first without a `values.yaml` among the 12 charts currently in `helm-repository`. Nothing upstream enforces either, and `helm lint --strict` passes, so this is cosmetic.

charts/nebari-app/templates/_nebari-app.tpl
```diff
@@ -0,0 +1,13 @@
+{{- define "nebari-app.nebariApp" -}}
+{{- $_ := required "metadata.name is required" .metadata.name -}}
+{{- $_ := required "metadata.namespace is required" .metadata.namespace -}}
```
dcmcand: This is stricter than the CRD. The generated schema has `required: ['spec']` at the root and `metadata: {type: object}` with no constraints, and the API server accepts a NebariApp with no `metadata.namespace` since the namespace comes from the request context.

Defensible in a Helm chart where `.Release.Namespace` is always available, and being explicit avoids surprises. Just noting it is a chart-level policy rather than an API requirement, in case a consumer ever wants to let `helm install -n` decide.

test/helm/nebari-app/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v2
+name: fixture
```
dcmcand, line 2: `test/` is otherwise Go only (`test/e2e/`, `test/utils/`). `test/helm/fixture/` or `charts/nebari-app/tests/` would signal what this is more clearly.

charts/nebari-app/README.md
```diff
@@ -0,0 +1,107 @@
...
+    "spec" (mergeOverwrite (dict
+        "service" (dict
+          "name" "frontend"
+          "port" .Values.frontend.service.port
+        )
+      ) .Values.frontend.nebariApp)
+) }}
+```
+
+`mergeOverwrite` gives precedence to its **second argument**, so consumer values override the computed defaults on overlapping keys; the computed `service.name`/`service.port` fill in the gaps.
```
dcmcand, line 74: Worth calling out that this is correct, including the deep-merge behaviour where the computed `service.name`/`service.port` survive if the consumer's `service` block omits them. `merge` versus `mergeOverwrite` argument precedence is an easy thing to state backwards, and getting it wrong in docs would send consumers down a confusing path.

charts/nebari-app/templates/_nebari-app.tpl
```diff
@@ -0,0 +1,13 @@
+{{- define "nebari-app.nebariApp" -}}
```
dcmcand, line 1: The pure-function shape is the right design for this: no `values.yaml`, no hidden coupling to the consumer's value tree, and callers stay in control of `metadata`. It also means the five guards line up exactly with the CRD's top-level required set (`spec.required = ['hostname', 'service']`, `service.required = ['name', 'port']`), with deeper validation deliberately left to the API server as README:54 says. Each guard is load-bearing: deleting any one of them makes `make helm-test` go red with the correct assertion firing.

.github/workflows/release.yml
```diff
@@ -200,3 +200,10 @@ jobs:
           chart-path: dist/chart
           chart-name: nebari-operator
           token: ${{ secrets.NEBARI_HELM_REPO_TOKEN }}
+
+      - name: Sync library chart to helm-repository
+        uses: nebari-dev/helm-repository/.github/actions/sync-chart@main
+        with:
+          chart-path: charts/nebari-app
+          chart-name: nebari-app
+          token: ${{ secrets.NEBARI_HELM_REPO_TOKEN }}
```
dcmcand, line 209: I checked this end to end because a dead-end here would only show up mid-release. Every file in the upstream sync, validate, auto-merge, publish, and signing path at `helm-repository@main` is type-agnostic: no step reads `Chart.yaml`'s `type`, requires a `values.yaml`, or runs `helm template`/`helm install`, and both `helm lint` calls omit `--strict`.

Then I ran the mechanical chain locally against this exact chart: `helm package`, `helm lint --strict`, `helm repo index` (preserves `type: library` in the entry), `helm show chart`, `helm push` to a local OCI registry, and a consumer chart resolving and rendering it back from OCI. All clean. `validate-sync-pr.yml` would return `pass` with zero warnings, so auto-merge would take it.

dcmcand (changes requested):
Nice piece of work. The library chart is a clean pure-function design, the three fixture cases mirror real consumer shapes, and using nebari-dev/chat-pack#162 to prove the API against a real chart rather than a toy was the right call.

Three blockers, one design question, and some smaller notes, all inline. Everything was verified by execution against `79a0595` or against upstream source at `nebari-dev/helm-repository@main`; commands and output are in the individual comments so you can re-run any of it.

**Blockers:** the dependency `repository:` URL 404s (`README.md:15`), the recommended `kubectl --dry-run=client` validation step silently passes invalid resources (`README.md:54`), and `make helm-test` stays green on a library chart that emits nothing at all (`Makefile:217-226`). The first two are one-line fixes. The third is a harness gap, not an output bug: the chart renders correctly today, I checked.

**Design question** on `Makefile:203-205`: the library chart's published version is currently slaved to the operator's release version. That is worth a deliberate decision, since two smaller issues fall out of it.

Two things I could not attach inline because the files are not in this diff:

- `docs/maintainers/release-process.md:139-141` enumerates release artifacts and lists only `nebari-operator-<version>.tgz`. Since `helm-package` now builds a second chart and the release step uploads `dist/*.tgz`, a second archive will appear. Nothing there is false, just incomplete, and `release-checklist.md:78,98` has no item to verify the `nebari-app` sync PR landed.
- Unrelated to this PR, but noticed while checking chart URLs: the root `README.md:115` advertises `oci://ghcr.io/nebari-dev/charts/nebari-operator`, while `docs/maintainers/release-process.md:153` and the upstream publish workflow both use `oci://quay.io/nebari/charts`. An anonymous ghcr probe returns 401 so I could not settle which is stale. Probably worth a follow-up issue.

One thing I verified specifically because it would have been expensive to discover at release time: the upstream sync, validate, auto-merge, publish, and signing path all accept a `type: library` chart with no `values.yaml`. Details in the comment on `release.yml:204`.

Makefile
pmeier: This is intended. Potentially there might be releases without any change for the app chart, but that is ok as it simplifies the mental model that the app chart matches the operator. So while it is potentially not required to update the app chart for an operator release it is the recommendation and avoids checking.

charts/nebari-app/templates/_nebari-app.tpl
pmeier: If its not required, we should remove it.

Makefile
pmeier: Isn't that the same for the operator chart as well? Not sure what do to about this as part of this PR

charts/nebari-app/README.md
pmeier: The library chart uses the exact same mechanism as the operator chart. Not sure what the ask is here. Fix the process in general or are you pointing out that I did something wrong here.

Makefile
dcmcand: Yes, you're correct. I am dropping this as a blocker since it is unrelated.

If you want it cheaply while you're in the file: the test-chart job already installs helm and kubebuilder and runs make helm-chart, so appending a make helm-package step there covers both charts in a couple of seconds with no cluster. Equally happy for this to be a follow-up issue or left alone. Your call, it's not blocking.

charts/nebari-app/README.md
dcmcand: Neither. You didn't do anything wrong, and I'm not asking for a process fix. My original comment was an observation and I never said what I actually wanted changed, so the confusion is on me. Here is the concrete ask, and I verified it against a real OCI registry after your reply.

You're right that the mechanism is the same as the operator chart. The difference is what the mechanism costs in each case. The operator chart gets installed with `helm install --version 1.0.0`, so a version in the docs is an example and the reader substitutes their own. The library chart gets consumed as a `dependencies:` entry, where that string is a resolution constraint. It fails the build instead of degrading.

Combined with the lockstep versioning you confirmed in the other thread, that breaks this snippet as soon as the chart ships. `release.yml:194` derives the version from the tag and `Makefile:203` stamps it into `charts/nebari-app/Chart.yaml`, so the next release publishes `nebari-app` at `0.1.0-alpha.21`. There is no `0.1.0` to resolve.

I pushed the chart to a local OCI registry at `0.1.0-alpha.21` and ran the README snippet verbatim:

```
# version: 0.1.0, repository: oci://<registry>/charts
$ helm dependency build ./consumer
Error: could not download oci://<registry>/charts/nebari-app:
  failed to resolve <registry>/charts/nebari-app:0.1.0: not found
```

The obvious correction fails too, for a different reason, and this is the one that caught the test fixture:

```
# version: ">=0.1.0"
Error: could not download oci://<registry>/charts/nebari-app:
  Could not locate a version matching provided version string >=0.1.0
```

Over OCI an exact version resolves as a tag, so `0.1.0` is just missing. A range goes through semver instead, and Helm excludes prereleases from any range that doesn't carry one itself. `>=0.1.0-0` clears both and resolves to `0.1.0-alpha.21`. I confirmed that over `oci://` and `file://`.

So the ask is one character in two places: `test/helm/nebari-app/Chart.yaml:9`, where `>=0.1.0` landed but needs the `-0`, and the README example here. When a final `v0.1.0` ships, an exact pin here becomes correct again if you'd rather switch back then.

dcmcand (commented):
Three remaining items from the re-review, all small and all verified locally at `c3478550`. Everything else from the last round is resolved.

Makefile
```diff
@@ -199,7 +200,41 @@ helm-chart-version: ## Update Helm chart version and appVersion (requires VERSIO
...
+helm-test: helm-lint ## Render the nebari-app chart for all cases and verify against golden files.
+	@command -v helm >/dev/null 2>&1 || { echo >&2 "helm is required but not installed. See https://helm.sh/docs/intro/install/"; exit 1; }
+	helm dependency build test/helm/nebari-app >/dev/null
+	@for c in static computed multi; do \
+		helm template t test/helm/nebari-app --set cases.$$c.enabled=true > /tmp/$$c.yaml; \
+		diff -q test/helm/nebari-app/golden/$$c.yaml /tmp/$$c.yaml >/dev/null \
+			|| { echo >&2 "case $$c: golden mismatch"; diff test/helm/nebari-app/golden/$$c.yaml /tmp/$$c.yaml; exit 1; }; \
+		echo "  case $$c: ok"; \
+	done
+	@rm -f /tmp/static.yaml /tmp/computed.yaml /tmp/multi.yaml
```
dcmcand, line 240: The golden files do their job, and I checked rather than assumed. I mutated the template four ways: bogus `apiVersion`, bogus `kind`, deleted `metadata:` block, and `nindent 2` to `nindent 0` on the spec. Every one came back as `case <c>: golden mismatch`, exit 2. That covers all eight mutations from my original comment, since every one of them is output-visible.

What's no longer covered is the guards themselves. Deleting all four `required` calls leaves the suite green:

```
$ grep -v 'required "' _nebari-app.tpl.bak > _nebari-app.tpl   # zero guards left
$ make helm-test
  case static: ok
  case computed: ok
  case multi: ok
✅ nebari-app chart tests passed
```

Golden files can't catch that by construction. All three cases supply every required field, so the render is byte-identical whether the guards are there or not. The old loop did go red when a guard was deleted, it just couldn't tell you which one. That was my complaint, and the fix for it was to assert the message rather than to drop the loop.

Since the guards are the entire validation surface of this chart, I'd put the loop back with the assertion added. I tested this version: green on intact code, and deleting a single guard fails naming that field. Placing it after the golden check also means an unrelated render break surfaces as a golden mismatch first, so the guard loop won't misattribute it.

```suggestion
	@for f in metadata.name spec.hostname spec.service.name spec.service.port; do \
		helm template t test/helm/nebari-app --set cases.static.enabled=true --set cases.static.$$f= 2>&1 >/dev/null \
			| grep -q "$$f is required" || { echo >&2 "expected required failure for $$f"; exit 1; }; \
		echo "  required guard $$f: ok"; \
	done
	@rm -f /tmp/static.yaml /tmp/computed.yaml /tmp/multi.yaml
```

Tabs in the recipe, as usual.

test/helm/nebari-app/Chart.yaml
```diff
@@ -0,0 +1,10 @@
...
+description: Test fixture consuming the nebari-app library chart.
+type: application
+version: 0.1.0
+appVersion: "0.1.0"
+dependencies:
+  - name: nebari-app
+    version: ">=0.1.0"
```
dcmcand: This is the other half of the ask from the README thread, and it's the same one character.

`>=0.1.0` reads like it should work, but Helm excludes prereleases from any range that doesn't carry one itself, so it won't match the `0.1.0-alpha.N` versions this chart actually publishes. Stamping in the version the next release will produce:

```
$ sed -i 's/^version:.*/version: 0.1.0-alpha.21/' charts/nebari-app/Chart.yaml
$ helm dependency build test/helm/nebari-app
Error: can't get a valid version for 1 subchart(s): "nebari-app"
  (repository "file://../../../charts/nebari-app", version ">=0.1.0")
```

CI never sees this, because it always runs on a clean checkout where the version is still `0.1.0`. It lands on whoever runs `make helm-chart-version` locally and then tries to run the tests, which is a confusing failure to debug when the thing you changed looks unrelated.

`>=0.1.0-0` resolves, and the full suite passes against the alpha. I confirmed the same constraint works over `oci://` too, so the README example can take the identical fix.

```suggestion
    version: ">=0.1.0-0"
```

Makefile
```diff
@@ -199,7 +200,41 @@ helm-chart-version: ## Update Helm chart version and appVersion (requires VERSIO
...
+	sed -i.bak "s/^appVersion:.*/appVersion: \"$(APP_VERSION)\"/" charts/nebari-app/Chart.yaml
+	rm -f charts/nebari-app/Chart.yaml.bak
+	@echo "✅ Updated chart versions to $(VERSION) and appVersion to $(APP_VERSION)"
+
+.PHONY: helm-lint
+helm-lint: ## Lint the helm charts
+	@command -v helm >/dev/null 2>&1 || { echo >&2 "helm is required but not installed. See https://helm.sh/docs/intro/install/"; exit 1; }
+	@if [ ! -d "dist/chart" ]; then echo "Error: dist/chart/ not found. Run 'make helm-chart' first."; exit 1; fi
+	helm lint dist/chart
+	helm lint charts/nebari-app
```
dcmcand: This check pulls the operator chart into the library chart's test path, because `helm-test` depends on `helm-lint`. `dist/` is gitignored, so on a fresh checkout:

```
$ make helm-test
Error: dist/chart/ not found. Run 'make helm-chart' first.
make: *** [Makefile:211: helm-lint] Error 1
```

That worked at `e554c68`. CI stays green because both `build-pr.yml` and `test-chart.yml` run `make helm-chart` first, so the cost lands entirely on contributors: testing a library chart that has nothing to do with the operator now means generating the operator chart, which means having kubebuilder installed.

Linting both charts in CI is the right call, so I'd keep that and split the target so the library chart can still be tested on its own:

```suggestion
.PHONY: helm-lint-app
helm-lint-app: ## Lint the nebari-app library chart.
	@command -v helm >/dev/null 2>&1 || { echo >&2 "helm is required but not installed. See https://helm.sh/docs/intro/install/"; exit 1; }
	helm lint charts/nebari-app

.PHONY: helm-lint
helm-lint: helm-lint-app ## Lint the helm charts.
	@if [ ! -d "dist/chart" ]; then echo "Error: dist/chart/ not found. Run 'make helm-chart' first."; exit 1; fi
	helm lint dist/chart
```

Then line 227 becomes `helm-test: helm-lint-app`, which GitHub won't let me fold into the same suggestion.

I ran this: `make helm-lint-app` passes with no `dist/`, `make helm-lint` still errors on the missing `dist/chart` as you intended, and `make helm-test` goes green on a checkout with no `dist/` at all.

charts/nebari-app/README.md
```diff
@@ -0,0 +1,106 @@
...
+The chart exposes one named template, `nebari-app.nebariApp`, that acts as a pure function: callers pass `metadata` and `spec` dicts and the template renders a `NebariApp` resource. It is not installable on its own — it is consumed as a [chart dependency](https://helm.sh/docs/helm/helm_dependency/).
+
+## Install as a dependency
+
+Add the chart to your consumer chart's `Chart.yaml`:
+
+```yaml
+dependencies:
+  - name: nebari-app
+    version: >=0.1.0-0
```
dcmcand, line 14: **Blocker.** Unquoted, this snippet is not valid YAML. A leading `>` is the folded-block-scalar indicator, so a reader who copies the fence verbatim gets a `Chart.yaml` Helm cannot load at all.

I extracted both fenced blocks byte-for-byte from this file rather than retyping them:

```
$ helm dependency build ./consumer
Error: cannot load Chart.yaml: error converting YAML to JSON:
  yaml: line 6: did not find expected comment or line break
```

Same failure, exit 1, for `dependency update`, `lint`, `template`, and `show chart`, and for both the `oci://` block here and the `file://` block at line 23. I pointed the second one at a valid local chart so resolution couldn't be the cause: it still fails, so this is purely parsing. PyYAML and `yq` reject it identically, so it's spec-level rather than a Helm quirk. The quoted form gives exit 0 and resolves.

Your `test/helm/nebari-app/Chart.yaml:9` already quotes it correctly, so this is only the copy-paste docs. Same one-character fix on line 23.

```suggestion
    version: ">=0.1.0-0"
```

.github/workflows/build-pr.yml
```diff
@@ -239,6 +239,9 @@ jobs:
       - name: Lint Helm Chart
         run: helm lint ./dist/chart
 
+      - name: Test nebari-app library chart
+        run: make helm-test
```
dcmcand, line 243: **Blocker, and this one is my fault.** Last round I asked you to move `make helm-test` out of the kind job into this job. I didn't check this workflow's path filter.

`on.pull_request.paths` at lines 7-18 has no `charts/**`, and none of the eleven patterns matches a `charts/nebari-app/**` path. GitHub evaluates `pull_request` path filters against the whole PR's three-dot diff, and if nothing matches, the workflow doesn't run. Confirmed against this repo's own history: PRs #75 (docs only), #82 (`PROJECT`), and #98 (`config/manager/manager.yaml`) all produced zero `PR Checks` check runs.

So a future PR editing only `_nebari-app.tpl` or the chart README won't trigger this workflow, and the step below never runs. `test-chart.yml` is unfiltered and still fires, but it runs `helm-chart`, `helm-lint`, and `helm-package` only, so no golden comparison and no guard assertions. Net effect: the change most likely to break the library chart is the one its tests can miss. At `79a0595` this was covered, because `make helm-test` sat in the unfiltered `test-chart.yml:59`; the move traded that away.

This PR itself is unaffected, since it touches `Makefile` and `test/**`.

One line at the end of the filter closes it:

```yaml
      - 'hack/**'
      - 'charts/**'
```

Makefile
```diff
@@ -199,7 +200,58 @@ helm-chart-version: ## Update Helm chart version and appVersion (requires VERSIO
...
+		helm template t test/helm/nebari-app --set cases.$$c.enabled=true > test/helm/nebari-app/golden/$$c.yaml; \
+		echo "  case $$c: golden written"; \
+	done
+	@echo "✅ Golden files written to test/helm/nebari-app/golden/"
+
+.PHONY: helm-test
+helm-test: helm-lint-library ## Render the nebari-app library chart for all cases and verify against golden files.
+	@command -v helm >/dev/null 2>&1 || { echo >&2 "helm is required but not installed. See https://helm.sh/docs/intro/install/"; exit 1; }
+	helm dependency build test/helm/nebari-app >/dev/null
+	@for c in static computed multi; do \
```
dcmcand, line 234: `test/helm/nebari-app/golden/minimal.yaml` is written but never read. The generate loop at line 224 covers `minimal static computed multi`; this compare loop covers only `static computed multi`. I overwrote the golden with `GARBAGE NOT YAML {{{ !!!` and `make helm-test` still exits 0, and `grep -rn golden` finds no other reader anywhere including `.github/`. The minimal *case* isn't dead, the guard loop renders it, but the file asserts nothing.

```suggestion
	@for c in minimal static computed multi; do \
```

test/helm/nebari-app/values.yaml
```diff
@@ -0,0 +1,91 @@
+cases:
+  minimal:
+    enabled: false
+    metadata:
+      name: "metatada.name"
+    spec:
+      hostname: "spec.hostname"
+      service:
+        name: "spec.service.name"
+        port: 0
```
dcmcand, line 10: Two small things in this fixture. `port: 0` is rejected by the CRD (`minimum: 1`, `config/crd/bases/reconcilers.nebari.dev_nebariapps.yaml:566`), and `required` passes int `0`, so nothing catches it. Harmless while the goldens are only string-compared, but it would block ever running them through `kubectl apply --dry-run=server`, which is what `README:53` recommends to consumers. Same for `hostname: spec.hostname`, which isn't a DNS name.

Also `metatada.name` on line 5 is a typo for `metadata.name`, mirrored into `golden/minimal.yaml:6`. Fittingly, that's the one file nothing asserts against.

dcmcand (changes requested):
Everything from the last round is done, and the guard loop is now genuinely load-bearing. I mutation-tested it rather than reading it: deleting each of the four `required` calls individually makes `make helm-test` exit 2 naming that exact field and stopping at the right point in the loop, deleting all four fails at `metadata.name`, and injecting an unrelated template error reachable only in the minimal case produces zero false `ok` lines. The `set +o pipefail` comment is accurate, `Makefile:19` really does set it. `make helm-test` also now runs green on a checkout with no `dist/` and no kubebuilder installed, which was the point of the split.

I also confirmed the `-0` is load-bearing rather than decoration: stamped `charts/nebari-app/Chart.yaml` to `0.1.0-alpha.21` (what `release.yml:149` produces), and `make helm-test` resolves to the alpha; the same run with `>=0.1.0` fails outright.

Two blockers left, both one line, both inline. Plus some smaller notes, none gating:

- `CONTRIBUTING.md:154-160` ("Adding a New Field to NebariApp") is six steps and doesn't mention the chart. The required-field set in `_nebari-app.tpl:2-5` hand-mirrors the CRD, and the drift that bites is narrowing: if `hostname` ever becomes optional, the chart hard-fails renders the API server would accept. A step 7 pointing at the guards and `make helm-test-generate-golden` is cheap.
- `helm-package:187` and `helm-chart-version:199` still gate on `dist/chart` even though `charts/nebari-app` is committed source with no dependency on the generated chart. The same split you just did for lint would apply.
- `helm-chart-version:203-205` sed-edits the committed `charts/nebari-app/Chart.yaml`, whereas every other release-time rewrite targets gitignored output. Harmless in CI, leaves a dirty tree for the emergency path at `release-process.md:312`.
- Still open from my first review: `release-process.md:140`, `:335`, and `release-checklist.md:84` list only `nebari-operator-<version>.tgz`, while `make helm-package` now emits a second archive (I ran it: `nebari-app-0.1.0.tgz`) and `release.yml:159` uploads `dist/*.tgz`. `grep -rn nebari-app docs/` returns nothing, so `docs/` doesn't know the chart exists.
- An ADR recording the scope change would be worth having, since `docs/decisions/` currently holds one ADR whose consequences section says the repo is "purely a Kubernetes controller".
- Pre-existing and separate: root `README.md:114` advertises `oci://ghcr.io/nebari-dev/charts/nebari-operator` while the maintainer docs and this chart use `oci://quay.io/nebari/charts`.

viniciusdc (commented):
Thanks for the follow ups @pmeier , Solid PR: the pure-`toYaml`-passthrough template is exactly the drift-immune shape #148 asked for, and the three blockers from the earlier comments are genuinely fixed on this head: 
- `charts/**` is in the `build-pr.yml` path filter, 
- the README dependency versions are quoted `">=0.1.0-0"`, 
- the golden compare loop covers `minimal`.

Two things I'd like to sort out on this PR: left inline on the specific lines. Everything else I found is minor, so I'll open follow-ups rather than hold this up.

viniciusdc (approved):
Approving to unblock. Chuck is out, and his three Aug-10 blockers are all addressed on 3719714 - I re-checked each (charts/** in the path filter, quoted dependency versions, golden minimal now compared).

The two items I raised - the ADR `Supersedes: ADR-001` line and the untested `mergeOverwrite` precedence - are non-blocking. I've split them plus a few smaller cleanups into follow-ups so they don't hold this up:

- #172 - harden the required-field guards (port 0 / missing service block)
- #173 - cover mergeOverwrite precedence + test-harness cleanups
- #174 - ADR-002 corrections (supersede, open questions, alternatives, overclaims)

We'll land those separately. Thanks for the follow-ups @pmeier.

PR #184, opened 2026-08-20
template values for library chart

pmeier:
## Issue

TL;DR

- NebariApp spec values like hostname or service configuration often need dynamically computed default values
- Computation can only happens as part of templates
- There is no good native way in helm to blend static and dynamic default values to pass them on as arguments to a template

## What does this implement/fix?

TL;DR

- reusable `deepTplJson` template that recursively applies the [`tpl` function](https://helm.sh/docs/v3/howto/charts_tips_and_tricks/#using-the-tpl-function) to all strings in a nested structure
- usage of `deepTplJson` for the metadata and spec of the `nebariApp` template to allow dynamically computed default values natively

---

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds a feature)
- [ ] Breaking change (fix or feature that would cause existing features not to work as expected)
- [x] Documentation Update
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build related changes
- [ ] Other (please describe):

## Testing

- [x] Did you test the pull request locally?
- [x] Did you add new tests?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -222,9 +222,9 @@ helm-lint: helm-lint-library ## Lint the helm charts
 .PHONY: helm-test-generate-golden
 helm-test-generate-golden: ## Generate golden files for nebari-app chart tests.
 	@command -v helm >/dev/null 2>&1 || { echo >&2 "helm is required but not installed. See https://helm.sh/docs/intro/install/"; exit 1; }
-	helm dependency build test/helm/nebari-app >/dev/null
+	helm dependency build --skip-refresh test/helm/nebari-app >/dev/null
 	mkdir -p test/helm/nebari-app/golden
-	@for c in minimal static multi; do \
+	@for c in minimal static multi templating; do \
 		helm template t test/helm/nebari-app --set cases.$$c.enabled=true > test/helm/nebari-app/golden/$$c.yaml; \
 		echo "  case $$c: golden written"; \
 	done
@@ -233,9 +233,9 @@ helm-test-generate-golden: ## Generate golden files for nebari-app chart tests.
 .PHONY: helm-test
 helm-test: helm-lint-library ## Render the nebari-app library chart for all cases and verify against golden files.
 	@command -v helm >/dev/null 2>&1 || { echo >&2 "helm is required but not installed. See https://helm.sh/docs/intro/install/"; exit 1; }
-	helm dependency build test/helm/nebari-app >/dev/null
+	helm dependency build --skip-refresh test/helm/nebari-app >/dev/null
 	@workdir=$$(mktemp -d); \
-	for c in minimal static multi; do \
+	for c in minimal static multi templating; do \
 		helm template t test/helm/nebari-app --set cases.$$c.enabled=true > $$workdir/$$c.yaml; \
 		diff -q test/helm/nebari-app/golden/$$c.yaml $$workdir/$$c.yaml >/dev/null \
 			|| { echo >&2 "case $$c: golden mismatch"; diff test/helm/nebari-app/golden/$$c.yaml $$workdir/$$c.yaml; rm -rf $$workdir; exit 1; }; \
--- a/charts/nebari-app/README.md
+++ b/charts/nebari-app/README.md
@@ -1,10 +1,6 @@
 # nebari-app
 
-A [library](https://helm.sh/docs/chart_template_guide/getting_started/#the-chart-yaml-file) Helm chart providing a reusable named template for rendering `NebariApp` custom resource instances.
-
-The chart exposes one named template, `nebari-app.nebariApp`, that acts as a pure function: callers pass `metadata` and `spec` dicts and the template renders a `NebariApp` resource. It is not installable on its own — it is consumed as a [chart dependency](https://helm.sh/docs/helm/helm_dependency/).
-
-> New here? For a step-by-step, end-to-end walkthrough (dependency → values → template → apply → verify), start with [Onboarding an app with the nebari-app Helm chart](../../docs/using-the-nebari-app-chart.md). This README is the reference for the template contract itself.
+A [library](https://helm.sh/docs/chart_template_guide/getting_started/#the-chart-yaml-file) Helm chart providing reusable templates for rendering `NebariApp` custom resource instances.
 
 ## Install as a dependency
 
@@ -13,7 +9,7 @@ Add the chart to your consumer chart's `Chart.yaml`:
 ```yaml
 dependencies:
   - name: nebari-app
-    version: ">=0.1.0-0"
+    version: ">=0.1.0"
     repository: oci://quay.io/nebari/charts
 ```
 
@@ -22,17 +18,70 @@ or via a local file path during development:
 ```yaml
 dependencies:
   - name: nebari-app
-    version: ">=0.1.0-0"
+    version: ">=0.1.0"
     repository: file://../nebari-app
 ```
 
 then run `helm dependency build`.
 
-## Usage
+## Templates
 
-The template takes a dict with `metadata` and `spec` keys:
+| Template | Description |
+|----------|-------------|
+| [`nebari-app.nebariApp`](#nebari-appnebariapp) | Renders a complete `NebariApp` resource from `metadata` and `spec` dicts |
+| [`nebari-app.deepTplJson`](#nebari-appdeeptpljson) | Applies template expansion to all strings in a nested structure, automatically parsing JSON outputs |
+
+> New here? For a step-by-step, end-to-end walkthrough (dependency → values → template → apply → verify), start with [Onboarding an app with the nebari-app Helm chart](../../docs/using-the-nebari-app-chart.md). This README is the reference for the template contracts.
+
+### nebari-app.nebariApp
+
+Renders a complete `NebariApp` custom resource.
+
+#### Usage
 
 ```yaml
+{{ include "nebari-app.nebariApp" (dict "metadata" $metadata "spec" $spec "ctx" $ctx) }}
+```
+
+#### Parameters
+
+- `$metadata`: Metadata mapping, such as name, namespace, and labels. Templates will be expanded using [`nebari-app.deepTplJson`](#nebari-appdeeptpljson).
+- `$spec`: `NebariApp` CR specification. Templates will be expanded using [`nebari-app.deepTplJson`](#nebari-appdeeptpljson).
+- `$ctx`: Optional templating context. If omitted, an empty context is used.
+
+#### Required fields
+
+The template enforces the presence of the following fields
+
+- `$metadata.name`
+- `$spec.hostname`
+- `$spec.service.name`
+- `$spec.service.port`
+
+as well as additionally the validity if the following fields
+
+- `$spec.service.port`
+
+All other validation (the rest of the `NebariAppSpec` schema) happens API-server-side at apply time. To catch schema errors before deployment, pipe `helm template` output through `kubectl apply --dry-run=server -f -`. This requires a cluster with the NebariApp CRD installed.
+
+#### Dynamic defaults
+
+The template uses `nebari-app.deepTplJson` internally to render both `$metadata` and `$spec`. This means you can embed template expressions directly in your values, and they will be expanded at render time:
+
+```yaml
+# values.yaml
+service:
+  port: 80
+
+nebariApp:
+  hostname: "{{ printf "%s.example.com" .Release.Name }}"
+  service:
+    name: "{{ printf "%s-service" .Release.Name }}"
+    port: "{{ .Values.service.port }}"
+```
+
+```yaml
+# template.yaml
 {{ include "nebari-app.nebariApp" (dict
     "metadata" (dict
       "name"      .Release.Name
@@ -40,42 +89,11 @@ The template takes a dict with `metadata` and `spec` keys:
       "labels"    (dict "app.kubernetes.io/name" .Chart.Name)
     )
     "spec" .Values.nebariApp
+    "ctx"  .
 ) }}
 ```
 
-### Required fields
-
-The template enforces these required fields via Helm's `required` function (render aborts if any is missing or empty):
-
-- `metadata.name`
-- `spec.hostname`
-- `spec.service.name`
-- `spec.service.port`
-
-All other validation (the rest of the `NebariAppSpec` schema) happens API-server-side at apply time. To catch schema errors before deployment, pipe `helm template` output through `kubectl apply --dry-run=server -f -`. This requires a cluster with the NebariApp CRD installed.
-
-### Dynamic service values
-
-When `spec.service.name` is derived (e.g. from a component name) rather than static, use `mergeOverwrite` to layer computed values under the consumer's static values:
-
-```yaml
-{{ include "nebari-app.nebariApp" (dict
-    "metadata" (dict
-      "name"      "frontend"
-      "namespace" .Release.Namespace
-    )
-    "spec" (mergeOverwrite (dict
-        "service" (dict
-          "name" "frontend"
-          "port" .Values.frontend.service.port
-        )
-      ) .Values.frontend.nebariApp)
-) }}
-```
-
-`mergeOverwrite` gives precedence to its **second argument**, so consumer values override the computed defaults on overlapping keys; the computed `service.name`/`service.port` fill in the gaps.
-
-### Multiple NebariApps from one chart
+#### Multiple NebariApps from one chart
 
 To avoid repeating `metadata` construction at every call site, define a thin consumer-owned wrapper template that builds `metadata` from `top` and `component` and forwards `spec`:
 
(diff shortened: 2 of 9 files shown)


charts/nebari-app/templates/_nebari-app.tpl
```diff
@@ -1,14 +1,17 @@
 {{- define "nebari-app.nebariApp" -}}
-{{- $_ := required "metadata.name is required" .metadata.name -}}
-{{- $_ := required "spec.hostname is required" .spec.hostname -}}
-{{- $_ := required "spec.service is required" .spec.service -}}
-{{- $_ := required "spec.service.name is required" .spec.service.name -}}
-{{- $_ := required "spec.service.port is required" .spec.service.port -}}
-{{- if lt (int .spec.service.port) 1 -}}{{- fail "spec.service.port must be >= 1" -}}{{- end -}}
+{{- $ctx := .ctx | default dict -}}
+{{- $metadata := include "nebari-app.deepTplJson" (dict "ctx" $ctx "value" .metadata) | fromJson -}}
+{{- $_ := required "metadata.name is required" $metadata.name -}}
```
pmeier: Move the checks after the templating as a template might expand into an empty value, while the template itself is not empty.

Makefile
```diff
@@ -222,9 +222,9 @@ helm-lint: helm-lint-library ## Lint the helm charts
 .PHONY: helm-test-generate-golden
 helm-test-generate-golden: ## Generate golden files for nebari-app chart tests.
 	@command -v helm >/dev/null 2>&1 || { echo >&2 "helm is required but not installed. See https://helm.sh/docs/intro/install/"; exit 1; }
-	helm dependency build test/helm/nebari-app >/dev/null
+	helm dependency build --skip-refresh test/helm/nebari-app >/dev/null
```
pmeier: Since the only dependency we have is local, we can skip the refresh of all registered repositories the user has, as they don't apply anyway.

charts/nebari-app/templates/_helpers.tpl
```diff
@@ -0,0 +1,61 @@
...
+    {{- $tplValue = append $tplValue (include "nebari-app.internal.deepTplJson" (dict "ctx" $ctx "value" $v) | fromJson).tplValue -}}
+{{- end -}}
+
+{{- /* strings: expand templates */ -}}
+{{- else if kindIs "string" $value -}}
+    {{- $tplValue = tpl $value $ctx -}}
+    {{- /* Try parse output as JSON and use it if valid. */ -}}
+    {{- /* This is required for templates to return anything but plain strings. */ -}}
+    {{- /* At the same time we cannot hard-require JSON output, as we are also handling regular static strings without a template */ -}}
+    {{- $tplValueFromJson := (printf "{\"tplValue\": %s}" $tplValue | fromJson).tplValue -}}
```
dcmcand: **Blocker.** This runs on every string, including ones that never contained a template, so any static value that parses as a JSON scalar gets retyped. `"1.0"` becomes the number `1`, `"8080"` becomes `8080`, `"true"` becomes a boolean, and `'{"a": 1}'` becomes a map.

That breaks string-typed fields. `ObjectMeta.Labels` is `map[string]string` in apimachinery, so a numeric label value is rejected at apply time, and the fixture in this PR hits it.

Guard on the source string actually containing a template before attempting the re-parse:

```
{{- if and (contains "{{" $value) ... }}
```

This does not fix the issue on the next line. They need separate fixes.

charts/nebari-app/templates/_helpers.tpl
```diff
@@ -0,0 +1,61 @@
...
+    {{- if $tplValueFromJson }}
```
dcmcand: **Blocker.** This gates on Go-template truthiness rather than on whether the parse succeeded, so any successfully-parsed falsy value gets thrown away and the string form survives. `{{ true }}` gives boolean `true`, `{{ false }}` gives the string `"false"`. Same asymmetry for `0` against `8080`, and for `[]`, `{}`, and `null`.

`spec.auth.enabled` is `type: boolean` in the CRD with no `x-kubernetes-preserve-unknown-fields` on `auth`, so `enabled: "false"` fails the structural schema. Nothing downstream can rescue it, since the API server rejects before the operator sees the object.

Gate on parse success instead. A sentinel in the probe object does it:

```
{{- $probe := printf "{\"ok\": true, \"tplValue\": %s}" $tplValue | fromJson -}}
{{- if $probe.ok }}
    {{- $tplValue = $probe.tplValue -}}
{{- end -}}
```

Taking this fix without the `{{` guard above widens the other bug, because static `"false"` and `"0"` currently survive only because they are falsy.

test/helm/nebari-app/golden/templating.yaml
```diff
@@ -0,0 +1,41 @@
+---
+# Source: fixture/templates/templating.yaml
+apiVersion: reconcilers.nebari.dev/v1
+kind: NebariApp
+metadata:
+  labels:
+    app.kubernetes.io/instance: t
+    app.kubernetes.io/name: t
+    app.kubernetes.io/version: 1
```
dcmcand: **Blocker**, same root as `_helpers.tpl:26`. The input is `"1.0"` and this records `1`, so the golden file locks in the coercion bug instead of catching it. `make helm-test` will now pass on a manifest the API server rejects.

This line should read `app.kubernetes.io/version: "1.0"` once the string branch is fixed.

test/helm/nebari-app/values.yaml
```diff
@@ -87,3 +87,38 @@ cases:
...
               - pathPrefix: /
                 pathType: PathPrefix
+  templating:
+    enabled: false
+    metadata:
+      name: "test-{{ .Release.Name }}-app"
+      namespace: default
+      labels:
+        app.kubernetes.io/name: "{{ .Release.Name }}"
+        app.kubernetes.io/version: "1.0"
```
dcmcand: Two cases worth adding here, which would have caught the other half: a template that renders boolean false, such as `enforceAtGateway: '{{ false }}'`, and one that renders zero.

charts/nebari-app/README.md
```diff
@@ -22,60 +18,82 @@ or via a local file path during development:
...
-When `spec.service.name` is derived (e.g. from a component name) rather than static, use `mergeOverwrite` to layer computed values under the consumer's static values:
+The template uses `nebari-app.deepTplJson` internally to render both `$metadata` and `$spec`. This means you can embed template expressions directly in your values, and they will be expanded at render time:
 
 ```yaml
+# values.yaml
+service:
+  port: 80
+
+nebariApp:
+  hostname: "{{ printf "%s.example.com" .Release.Name }}"
```
dcmcand: **Blocker.** The double-quoted scalar terminates at the inner `"`, so this block is invalid YAML. PyYAML and `helm template` both fail with `found character '%' that cannot start any token`. Line 79 has the same problem.

Outer single quotes fix it, which is what the fixture already does at `test/helm/nebari-app/values.yaml:112`:

```yaml
hostname: '{{ printf "%s.example.com" .Release.Name }}'
```

charts/nebari-app/README.md
```diff
@@ -22,60 +18,82 @@ or via a local file path during development:
...
-- `spec.service.name`
-- `spec.service.port`
+The template enforces the presence of the following fields
+
+- `$metadata.name`
+- `$spec.hostname`
+- `$spec.service.name`
+- `$spec.service.port`
+
+as well as additionally the validity if the following fields
```
dcmcand: `$spec.service.port` appears in both lists, and the connective sentence doesn't quite parse. Suggest one list plus one sentence:

> The template enforces the presence of `$metadata.name`, `$spec.hostname`, `$spec.service.name`, and `$spec.service.port`, and rejects a port below 1.

charts/nebari-app/templates/_nebari-app.tpl
```diff
@@ -1,14 +1,31 @@
+{{/*
+  Render a complete `NebariApp` custom resource.
+
+  Usage:
+    {{ include "nebari-app.nebariApp" (dict "metadata" $metadata "spec" $spec "ctx" $ctx) }}
+
+  Parameters:
+    - `$metadata`: Metadata mapping, such as name, namespace, and labels. Templates will be expanded using ebari-app.deepTplJson.
```
dcmcand: Typo: `ebari-app.deepTplJson`.

charts/nebari-app/templates/_nebari-app.tpl
```diff
@@ -1,14 +1,31 @@
...
 {{- define "nebari-app.nebariApp" -}}
-{{- $_ := required "metadata.name is required" .metadata.name -}}
-{{- $_ := required "spec.hostname is required" .spec.hostname -}}
-{{- $_ := required "spec.service is required" .spec.service -}}
-{{- $_ := required "spec.service.name is required" .spec.service.name -}}
-{{- $_ := required "spec.service.port is required" .spec.service.port -}}
-{{- if lt (int .spec.service.port) 1 -}}{{- fail "spec.service.port must be >= 1" -}}{{- end -}}
+{{- $ctx := .ctx | default dict -}}
+{{- $metadata := include "nebari-app.deepTplJson" (dict "ctx" $ctx "value" .metadata) | fromJson -}}
+{{- $_ := required "metadata.name is required" $metadata.name -}}
```
dcmcand: Moving the `required` checks below the templating is the right call. Worth a one-line comment saying why: a non-empty template can expand to an empty value, so checking the raw input would pass where the rendered output should fail.

dcmcand (changes requested):
Thanks for this Philip, this is looking good. The `mergeOverwrite` approach only ever handled whole-value replacement, so `deepTplJson` is the right shape for the general case, and the README reorganization is a real improvement.

Two blockers in code and one in the docs, all inline. The code ones are both in the string branch of `deepTplJson` and they need separate fixes: I tested each in isolation and neither one fixes the other. Fixing only the truthiness gate makes the static-string case worse.

Three smaller notes that don't fit inline:

- Every value string now goes through `tpl`, including when the caller omits `ctx`, so a value containing a literal `{{` aborts the render where it used to pass through. It fails loudly at `helm template` time, so I'd just document the escape rather than change the code, but it belongs in the release notes since the PR is marked non-breaking.
- `docs/using-the-nebari-app-chart.md:12` still says the chart exposes one named template, and that walkthrough never mentions `ctx`, so someone following it end to end can't use what this adds.
- `mergeOverwrite` loses all its docs here, but `multi.yaml` still uses it and the fixture comment still explains its precedence. Either keep a sentence on it or deprecate it and convert the fixture.

charts/nebari-app/templates/_helpers.tpl
```diff
@@ -0,0 +1,58 @@
...
+
+{{- /* slices: recursively handle elements */ -}}
+{{- else if kindIs "slice" $value -}}
+    {{- $tplValue = list -}}
+    {{- range $v := $value -}}
+        {{- $tplValue = append $tplValue (include "nebari-app.internal.deepTplJson" (dict "ctx" $ctx "value" $v) | fromJson).tplValue -}}
+    {{- end -}}
+
+{{- /* templates: render */ -}}
+{{- else if and (kindIs "string" $value) (contains "{{" $value) -}}
```
dcmcand: Question, non-blocking, and the one I care most about.

Was strict must-render-valid-JSON a deliberate choice over "parse if valid JSON, else keep the string"? The `contains "{{"` guard makes the lenient fallback much safer than it was last round, and it would let `'{{ .Release.Name }}.example.com'` work without `| toJson`.

The reason I ask: the strict rule fails loudly only when the output *isn't* valid JSON. When it happens to be, the value is silently retyped. `'{{ .Chart.AppVersion }}'` with `appVersion: "1.0"` renders `app.kubernetes.io/version: 1`, `helm template` succeeds, and the API server rejects it -- `ObjectMeta.Labels` is `map[string]string`, and both `json.Unmarshal` into typed `ObjectMeta` and `runtime.DefaultUnstructuredConverter.FromUnstructured` refuse the number. Annotations do the same thing.

I documented that as a hazard in the commit I pushed, but documenting it is the weaker fix. The stronger one is to exempt `$metadata` from JSON parsing entirely: `name`, `namespace`, `labels`, and `annotations` have no non-string leaf, so parsing there can never produce a valid value. Happy for that to be a follow-up rather than this PR -- but if you'd rather go lenient instead, say so and I'll redo the docs paragraph, because it describes the strict contract.

charts/nebari-app/templates/_nebari-app.tpl
```diff
@@ -1,14 +1,39 @@
...
 {{- define "nebari-app.nebariApp" -}}
-{{- $_ := required "metadata.name is required" .metadata.name -}}
-{{- $_ := required "spec.hostname is required" .spec.hostname -}}
-{{- $_ := required "spec.service is required" .spec.service -}}
-{{- $_ := required "spec.service.name is required" .spec.service.name -}}
-{{- $_ := required "spec.service.port is required" .spec.service.port -}}
-{{- if lt (int .spec.service.port) 1 -}}{{- fail "spec.service.port must be >= 1" -}}{{- end -}}
+{{- /* Template first, then validate: a non-empty template can expand to an empty value, so checking the raw input would pass where the rendered output should fail. */ -}}
+{{- $metadata := .metadata -}}
+{{- if .tplCtx -}}
```
dcmcand: Question, non-blocking. This is a truthiness test, so passing an empty `dict` as `tplCtx` silently skips templating rather than acting as an empty context, and the literal template then lands in the CR:

```yaml
spec:
  hostname: '{{ printf "%s.example.com" .Release.Name | toJson }}'
```

Intended, or would `hasKey . "tplCtx"` be better? I genuinely don't think either is obviously right -- an empty ctx would make every `.Release.Name` reference fail, which isn't clearly friendlier than skipping. Mostly flagging that the failure mode is silent.

charts/nebari-app/templates/_helpers.tpl
```diff
@@ -0,0 +1,58 @@
...
+    {{- $probe := printf "{\"ok\": true, \"tplValue\": %s}" (tpl $value $ctx) | fromJson -}}
```
dcmcand: This is the part I'd point someone at. Wrapping the parse in a sentinel object separates "parsed to a falsy value" from "didn't parse at all", which is exactly what the previous version couldn't express, and naming the offending template in the failure message means the error points at the values file instead of at the chart internals.

test/helm/nebari-app/values.yaml
```diff
@@ -45,45 +45,77 @@ cases:
...
+    spec:
+      hostname: '{{ printf "%s.nebari.local" .Release.Name | toJson }}'
+      service:
+        name: '{{ printf "svc-%s" .Release.Name | toJson }}'
+        port: '{{ .Values.cases.minimal.spec.service.port }}'
+      auth:
+        enabled: true
+        provider: keycloak
+        provisionClient: true
+        enforceAtGateway: '{{ false }}'
```
dcmcand: `'{{ false }}'` here and `'{{ 0 }}'` at line 118 are what make the golden load-bearing. I reverted the parse-success gate to check, and these are precisely the two lines that go red -- without them the golden would have passed on the old behavior.

dcmcand (approved):
Approving. All four blockers from the last round are fixed, and I verified each by mutation rather than by reading: dropping the `contains "{{"` guard turns `case templating` red, and reverting to the truthiness gate reproduces `enforceAtGateway: "false"` / `priority: "0"` exactly. The `$probe.ok` sentinel is the right shape for separating "parsed to a falsy value" from "didn't parse", and the `tplCtx` opt-in genuinely restores backward compatibility -- `main` has no templating at all, and omitting `tplCtx` reproduces that.

**Heads up: I pushed a docs commit to this branch** (`1afeb14`) rather than leaving you eight suggestions to click. It fixes things that were broken rather than debatable:

- The `Dynamic defaults` example in the chart README and the `Templating in values` example in the walkthrough both rendered bare strings, which are not valid JSON, so they hit the new `fail`. I built a consumer chart from each block verbatim to confirm, and again after the fix. Both now pipe through `| toJson`.
- README.md:68 had a stray U+00C4 where the closing `'` should be, so that block was invalid YAML.
- The `deepTplJson` description still said "all strings" in three places, which stopped being true when the `contains "{{"` guard went in. That guard is the reason `"1.0"` survives as a string, so it was worth stating.
- Noted that the chart now exposes two public templates, and added a line saying why the required checks sit below the templating.

Revert or reword any of it freely -- it's your PR and none of it is load-bearing. `make helm-test` is green, golden output is byte-identical, and CI should confirm.

Two questions inline. Neither blocks the merge; the second one is the one I'd actually like your read on, since it decides whether a sharp edge stays in the contract.

charts/nebari-app/templates/_helpers.tpl
pmeier: > `name`, `namespace`, `labels`, and `annotations` have no non-string leaf, so parsing there can never produce a valid value.

This sentence makes no sense to me. Please give me an example what you are concerned about.

charts/nebari-app/templates/_nebari-app.tpl
pmeier: `hasKey` makes sense.

charts/nebari-app/templates/_helpers.tpl
pmeier: I'd rather leave the behavior strict for now, because

1. It gives a clear contract. Everything has to be JSON and strings are not an exception.
2. It avoids a potentially complex handling and testing.

I'm going to check 2. and only if that comes back as "not complex at all" I'm open to relax 1.

pmeier: The issue is silent failure: if a user supplies a string that should contain valid JSON but makes a mistake, we have no way to distinguish that from an intended plain string. If we are lenient, this would pass silently and the user would now get a broken JSON string in place where they expected a properly rendered template. IMO this is worse than failing loudly. Pending https://github.com/nebari-dev/nebari-operator/pull/184#discussion_r3878743988 I'm going to resolve this.

test/helm/nebari-app/golden/multi.yaml
pmeier: The changes in this file are caused by the switch from `toYaml` to `toYamlPretty`. I like the pretty version better, but I can also revert. LMK if you want that. Comment applies to the static.yaml golden reference as well.

pmeier:
I've decided to disable templating in the metadata input for now. These values in most cases are set as part of a template anyway and thus differ significantly from the spec, which is usually pass through from the values and thus requires templating for dynamic defaults.

